PR #14, opened 2020-05-08
Adding a related Minecraft environment

MadcowD:
MineRL is a successor to Malmo and it would be wonderful if you all would be willing to include it here.


rockt:
Thanks William!

PR #18, opened 2020-05-12
Disable E402 for imports in agent.py

edran:
Could have moved the whole env + torch import bit down instead, but there's a chance we set the imports this way to avoid import bugs wrt gym and torch, so better to be conservative with these changes until we can confirm otherwise.


edran:
Waiting for tests to pass before merging.

PR #25, opened 2020-05-13
Create NLE hint files, and restore originals to NH 3.6.6

edran:
This is done to allow us to clean up these hints without merge issues (in another PR :) ).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -65,7 +65,7 @@ commands:
           command: |
             conda activate nle
             cd sys/unix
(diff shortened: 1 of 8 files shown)


heiner:
BTW I've had some success moving NetHack's build system to cmake: https://github.com/facebookresearch/nle/commit/6d87b69405dcc570755906b5afa98c2480266656 (this doesn't quite work 100% on that branch yet due to Lua, but should probably work for NetHack 3.6).

PR #26, opened 2020-05-13
Clean up NLE hint files

edran:


edran:
Closing, since we are switching to cmake.

Issue #28, opened 2020-05-25
Can't install on MacOS


TrentBrick:
## 🐛 Bug

I followed all of the install instructions and am getting the following error when I run `pip install nle`

```
Building wheel for nle (PEP 517) ... error
  ERROR: Command errored out with exit status 255:
   command: /Users/trentonbricken/anaconda3/envs/nle/bin/python /Users/trentonbricken/anaconda3/envs/nle/lib/python3.8/site-packages/pip/_vendor/pep517/_in_process.py build_wheel /var/folders/1q/6pbw30m128z5zbp5kfft7rn40000gn/T/tmps216ak48
       cwd: /private/var/folders/1q/6pbw30m128z5zbp5kfft7rn40000gn/T/pip-install-6l1fgazr/nle
  Complete output (46 lines):
  fatal: not a git repository (or any of the parent directories): .git
  Building wheel nle-0.1.1
  Successfully built fbs files from win/rl
  Found existing nle/fbs -- Removing old package
  Copying win/rl/nle/fbs to nle/fbs
  running bdist_wheel
  running build
  Building and installing NetHack...
  ./mkmkfile.sh: line 14: ../../doc/Makefile: No such file or directory
  ./mkmkfile.sh: line 15: ../../doc/Makefile: No such file or directory
  ./mkmkfile.sh: line 16: ../../doc/Makefile: No such file or directory
  ./mkmkfile.sh: line 17: ../../doc/Makefile: No such file or directory
  ./mkmkfile.sh: line 18: ../../doc/Makefile: No such file or directory
  ./mkmkfile.sh: line 19: ../../doc/Makefile: No such file or directory
  ./mkmkfile.sh: line 21: ../../doc/Makefile: No such file or directory
  ./mkmkfile.sh: line 22: ../../doc/Makefile: No such file or directory
  ./mkmkfile.sh: line 23: ../../doc/Makefile: No such file or directory
  ./mkmkfile.sh: line 26: ../../doc/Makefile: No such file or directory
  ./mkmkfile.sh: line 27: ../../doc/Makefile: No such file or directory
  ./mkmkfile.sh: line 28: ../../doc/Makefile: No such file or directory
  ./mkmkfile.sh: line 30: ../../doc/Makefile: No such file or directory
  ./mkmkfile.sh: line 31: ../../doc/Makefile: No such file or directory
  ./mkmkfile.sh: line 32: ../../doc/Makefile: No such file or directory
[...]

PR #32, opened 2020-06-09
Set clang as linker on MacOS

edran:
This is to further patch out #28 for today's release.


sys/unix/hints/macosx-nle
```diff
@@ -81,7 +81,7 @@ endif
 ifdef WANT_WIN_RL
 #CFLAGS += -URL_GRAPHICS -DRL_GRAPHICS -DNOUSER_SOUNDS
 CXXFLAGS += -std=c++14 -stdlib=libc++ -mmacosx-version-min=10.7
-LINK=g++
+LINK=clang
```
heiner, line 84: I'd have assumed we'd need `clang++` here but if this works that's fine too. :)

edran: No, you are right, it's a fair point. I should stop abusing clang's cleverness.

Fixed.

PR #38, opened 2020-06-24
Implement cmake-based build

edran:
## TODOs
- [x] ~Check all build flags for util/ and dat/~ I'm 80% sure we are good here.
- [x] Move generated files in dat/ into build/dat
- [x] Implement install properly
- [x] Declare target & install for util/recover
- [x] Refactor c++ build stuff for win/rl into its own cmakelists
- [x] Separate includes and libs sensibly
- [x] Test prefix behaviour with venv
- [x] Test prefix behaviour with conda
- [x] Switch python build to use cmake


dat/CMakeLists.txt
```diff
@@ -0,0 +1,277 @@
...
+
+# TODO(NN): try to glob .lev automatically
+#
+# NOTE: Using:
+#
+# install(DIRECTORY ${NLE_DAT_GEN} DESTINATION ${INSTDIR} FILES_MATCHING PATTERN
+# "*.lev")
+#
+# results in `$HACKDIR/dat/*.lev` due to CWD. I've tried to get it to work, but
+# given up after a couple of hours :(
```
edran, line 148: @heiner FYI. :(

edran: More context: ideally we want this list to be generated when running make install. This is possible using `install(DIRECTORY ... FILES_MATCHING PATTERN)`, but the resulting install paths are relative to `build/` dir, rather than `build/dat`.

dat/CMakeLists.txt
```diff
@@ -0,0 +1,276 @@
...
+    Ranger.des
+    Rogue.des
+    Samurai.des
+    Tourist.des
+    Valkyrie.des
+    Wizard.des)
+list(TRANSFORM QUEST_LEVELS PREPEND ${NLE_DAT}/)
+
+add_custom_command(
+  DEPENDS makedefs ${NLE_DAT}/bogusmon.txt ${NLE_DAT}/engrave.txt
```
heiner, line 32: What's NLE_DAT doing here? Shouldn't this "just work" with e.g. `bogusmon.txt` b/c it's in the same folder?

edran: Sorry, I still had to refactor this (and forgot to update the todo list).

This version of the code sort of relies on BINARY_DIR to be a subdirectory of SOURCE_DIR (but I was a little bit more defensive in writing the paths for the dependencies). In insight, I should be fixing the original problem. Doing this now.

edran: Oh wait, if you are referring to -- particularly -- the depends lines, no, that's with respect to BINARY_DIR (so we are in e.g. `build/dat`).

edran:
>I think this could be easier if we included pybind11 in the repo (or used find_package). Example: https://github.com/facebookresearch/nle/blob/libnethack/CMakeLists.txt#L52

I considered doing this, but imo it's not a great design choice from the pybind11 side... I think it's neater if we rely on the pip package and we check the status of the python command in cmake. We also need to guard these commands on pseudo target, otherwise they will always be run on `make` (and we don't necessarily want that).

edran:
@heiner FYI I'm going to re-request your review after I make the PR ready and working.

win/rl/CMakeLists.txt
```diff
@@ -0,0 +1,58 @@
...
+# target_link_libraries(helper PRIVATE $<TARGET_OBJECTS:nlehack>)
+
+add_library(monst OBJECT ${NLE_SRC}/monst.c)
+target_include_directories(monst PRIVATE ${NLE_INC} ${NLE_INC_GEN})
+add_library(decl OBJECT ${NLE_SRC}/decl.c)
+target_include_directories(decl PRIVATE ${NLE_INC} ${NLE_INC_GEN})
+add_library(drawing OBJECT ${NLE_SRC}/drawing.c)
+target_include_directories(drawing PRIVATE ${NLE_INC} ${NLE_INC_GEN})
+
+target_link_libraries(helper PUBLIC monst decl drawing)
```
edran, line 41: @heiner I caved in and added pybind11 as a dependency.

I'm a little stuck on trying to link the external objects to helper. Can you give it a look? I've tried various things, but all seem to leave undefined symbols in the `helper.so`.

heiner: You'll need to link statically.

edran: Indeed, but I tried that for a while -- groan, turns out adding `POSITION_INDEPENDENT_CODE ON` as a property to the libs does it.

nle/nethack/nethack.py
```diff
@@ -45,24 +45,9 @@
...
-        while True:
-            line = f.readline()
-            if not line:
-                raise ValueError("Could not determine HACKDIR from installed nethack")
-            if line.startswith("HACKDIR="):
-                HACKDIR = line[8:-1]  # Exclude newline.
-                break
-
-
+HACKDIR = os.getenv("HACKDIR", os.path.expanduser("~/nethackdir"))
```
heiner, line 48: We shouldn't make this assumption in in end, right?

edran: Indeed, I'm going to change this after I get the tests up and running :)

edran:
@heiner This is ready to review. Tests all pass now :)

There are a couple of things that still need to be handled:
1. Installing `nethackdir` into the repository (I tried a little bit and couldn't find a robust solution yet)
2. Remove generated .lev files, which make nethackdir a little bit too big. We are now generating nhdat (which was a major pain to handle, I had to refactor everything yesterday), so this should be an easy fix.

nle/nethack/nethack.py
heiner: Wait this is still in here?

util/CMakeLists.txt
```diff
@@ -0,0 +1,92 @@
...
+  COMMAND bison -y --defines=${NLE_UTIL_GEN}/lev_comp.h -o
+          ${NLE_UTIL_GEN}/lev_parser.c ${NLE_UTIL}/lev_comp.y
+  DEPENDS ${NLE_UTIL}/lev_comp.y)
+
+add_custom_command(
+  OUTPUT ${NLE_UTIL_GEN}/lev_scanner.c
+  COMMAND flex -o${NLE_UTIL_GEN}/lev_scanner.c ${NLE_UTIL}/lev_comp.l
+  DEPENDS ${NLE_UTIL}/lev_comp.l)
+
+add_custom_target(nleutil DEPENDS ${MAKEDEFS_HEADERS} recover)
```
heiner, line 86: Why call this `nleutil`?

nle/nethack/nethack.py
edran: Sorry, I left this as-is since the changes for **robustly** installing `$HACKDIR` into the package require more work. Why is this such a problem right now? The end behaviour is mostly similar to `master`, since we set it [here](https://github.com/facebookresearch/nle/pull/38/files#diff-af3b638bc2a3e6c650974192a53c7291R13-R15)?

util/CMakeLists.txt
edran: Good point -- I'll rename it util (and `nledat -> dat`).

nle/nethack/nethack.py
heiner: No, this is actually worse than before. Before it was a directory in conda. How it interferes with the users NetHack version (potentially overriding save games on compilation).

edran: Okay, I've changed this to `sys.base_prefix/nethackdir`, which is basically what we had before.

PR #45, opened 2020-07-03
Install HACKDIR into package

edran:
This PR modifies `setup.py` to install `HACKDIR` into the package, allowing to use `pkg_resources` to find it at runtime, as well as allowing `pip` to remove it when uninstalling the package.

When installing the package in editable mode, `HACKDIR` will be equal to `nle/nethackdir`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ $ pip install nle
 NOTE: If you want to extend / develop NLE, please install the package as follows:
 
 ``` bash
-$ git clone git@github.com:facebookresearch/nle
+$ git clone https://github.com/facebookresearch/nle --recursive
 $ pip install -e ".[dev]"
 $ pre-commit install
 ```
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -82,7 +82,7 @@ $ pip install nle
 NOTE: If you want to extend / develop NLE, please install the package as follows:
 
 ``` bash
-$ git clone git@github.com:facebookresearch/nle
+$ git clone https://github.com/facebookresearch/nle --recursive
```
heiner, line 85: Do we have submodules right now?

edran: Yes, in the earlier PR we added pybind11 in `third_party/pybind11`: https://github.com/facebookresearch/nle/tree/master/third_party

edran:
>Cool if this works!

Cheers -- CI currently seems to be OK with the changes, and I tested locally on both Ubuntu bionic and Ubuntu xenial.

PR #52, opened 2020-07-20
Fix lodash version due to GHSA-p6mc-m468-83gw.

heiner:
The gift that keeps giving. Why do we require dozens of npm packages?

Reference(s):
https://www.npmjs.com/advisories/1523
https://github.com/advisories/GHSA-p6mc-m468-83gw


edran:
We don't explicitly make use of them, but the few dependencies that we have pull tons of npm packages (like most of its ecosystem's packages) :(

PR #54, opened 2020-07-23
Require cmake version >= 3.14

heiner:
Upgrading from cmake version 3.13.1 made this error go away:

CMake Error at CMakeLists.txt:67 (install):
      install TARGETS given no RUNTIME DESTINATION for executable target
      "nethack".


heiner:
> > Easy as pi!
> 
> 😃
> 
> Regarding the PR, 3.14 is indeed the correct version to ensure compatibility. Sorry about that. CMake is supposed to warn us when there's a mismatch between minimum version and incompatible function calls, but this is probably a weird corner case and I didn't catch it myself either :(

No worries, and thanks for spotting my hidden joke.

Issue #21, opened 2020-05-12
Glyph ID handling


heiner:
Currently, each dungeon tile (ignoring the char/color/specials observation that's also available) is an int16 between `0` and `nethack.MAX_GLYPH == 5976`. We use an embedding lookup table of that size `embedding_dim == 32`. That's `5976 * 32 == 191232` floating points, or `191232 * 16 == 3059712` bits, or ~0.3MB. That doesn't seem too much but there's some [issues with the embedding itself](https://github.com/pytorch/pytorch/issues/24912). Also, it does not give the agent a cue that certain ids (e.g., dog and large dog) are more related than others (large dog vs wall).

The way these glyphs are organized is that first come all the monsters (`NUMMONS` many, which is 381), then pets (again `NUMMONS` many because in theory every monster can be tame, then a single glyph for an invisible monster (`GLYPH_INVIS_OFF`, which is 762), then a glyph for each "detected" monster (again `NUMMONS` many). For some obscure reason, then there's corpses, which are not monsters (but there's `NUMMONS` many), and then there's ridden monsters, which are monsters (`NUMMONS` many). The check `glyph_is_monster(glyph)` does [this](https://github.com/fairinternal/NetHack/blob/rl/include/display.h#L407):

```
#define glyph_is_monster(glyph)                            \
    (glyph_is_normal_monster(glyph) || glyph_is_pet(glyph) \
     || glyph_is_ridden_monster(glyph) || glyph_is_detected_monster(glyph))
```

This makes a list like `[i for i in range(nethack.MAX_GLYPH) if nethack.glyph_is_monster(i)]` have length `nethack.NUMMONS*4 == 1524`, but it's not contiguous.

Cf. https://github.com/fairinternal/NetHack/blob/rl/win/rl/helper.cc#L37 for a list of the offsets and take a look at the comment in https://github.com/fairinternal/NetHack/blob/rl/include/display.h#L235 explaining this.

After monsters there's `MAXPCHARS == 96` `cmap` entries for dungeon features, then there's zap beams (`NUM_ZAP << 2 == 8 << 2 == 32` many). Then there's `NUMMONS << 3 == 3048` (!) "swallow" glyphs. That's a lot for stuff that basically never happens to our agents. Then there's `WARNCOUNT == 6` warning glyphs and finally `NUMMONS` statue glyphs.

As a graphic representation, the glyph ids are:

```
[...]

PR #55, opened 2020-07-29
Update helpers

heiner:
This is to have a slightly better handle on #21.

### DIFF
--- a/win/rl/helper.cc
+++ b/win/rl/helper.cc
@@ -33,6 +33,7 @@ PYBIND11_MODULE(helper, m)
     m.attr("MAXWIN") = py::int_(MAXWIN);
 
     m.attr("NUMMONS") = py::int_(NUMMONS);
+    m.attr("NUM_OBJECTS") = py::int_(NUM_OBJECTS);
 
     // Glyph array offsets. This is what the glyph_is_* functions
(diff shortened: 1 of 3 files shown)


win/rl/helper.cc
```diff
@@ -151,6 +157,10 @@ PYBIND11_MODULE(helper, m)
         ;
 
     py::class_<class_sym>(m, "class_sym")
+        .def_static(
+            "from_mlet",
+            [](char let) -> const class_sym * { return &def_monsyms[let]; },
+            py::return_value_policy::reference)
```
edran, line 163: This is... interesting. Was this to bypass const?

edran: NVM, I saw the comment in the commit.

heiner: This is a bit of an unusual situation for pybind: NetHack maintains an array of class_sym structs. We invent a Python class for class_syms, but we don't actually want to create (or modify) any class_sym objects, just expose the entries of the array. Returning a pointer and using py::return_value_policy::reference makes pybind not take ownership.

What doesn't work is turning this into __init__ with the py::nodelete trick as in permonst (same situation otherwise). So we'll just use this static method.

PR #56, opened 2020-07-29
Enable chars, colors, specials as observation tensors.

heiner:
This adds "chars", "colors", and "specials" as observation keys. The latter is an xor'ed version of

```c
#define MG_CORPSE  0x01
#define MG_INVIS   0x02
#define MG_DETECT  0x04
#define MG_PET     0x08
#define MG_RIDDEN  0x10
#define MG_STATUE  0x20
#define MG_OBJPILE 0x40  /* more than one stack of objects */
#define MG_BW_LAVA 0x80  /* 'black & white lava': highlight lava if it
                            can't be distringuished from water by color */
```
(see hack.h)

This PR adds them as individual uint8 tensors. We could also return something of shape [3, DUNGEON_HEIGHT, DUNGEON_WIDTH]?

### DIFF
--- a/nle/env/base.py
+++ b/nle/env/base.py
@@ -81,14 +81,30 @@ def _get(response, path="Blstats.score", default=None, required=False):
     return node
 
 
-def _get_glyphs(response):
+def _get_as_np(response, attr, shape, dtype):
     if response is None:
-        return np.zeros(DUNGEON_SHAPE, dtype=np.int16)
+        return np.zeros(shape, dtype=dtype)
     o = response.Observation()
     # If done is True, Observation() is None.
     if o is None:
-        return np.zeros(DUNGEON_SHAPE, dtype=np.int16)
-    return o.Glyphs().DataAsNumpy().view(np.int16).reshape(DUNGEON_SHAPE)
+        return np.zeros(shape, dtype=dtype)
+    return getattr(o, attr)().DataAsNumpy().view(dtype).reshape(shape)
+
+
+def _get_glyphs(response):
+    return _get_as_np(response, "Glyphs", DUNGEON_SHAPE, np.int16)
+
+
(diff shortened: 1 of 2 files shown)


rockt:
I think three different tensors is fine for now. If it turns out that all three of them are important to use, we can change this to one tensor.

nle/env/base.py
```diff
@@ -81,14 +81,30 @@ def _get(response, path="Blstats.score", default=None, required=False):
...
     if response is None:
-        return np.zeros(DUNGEON_SHAPE, dtype=np.int16)
+        return np.zeros(shape, dtype=dtype)
     o = response.Observation()
     # If done is True, Observation() is None.
     if o is None:
-        return np.zeros(DUNGEON_SHAPE, dtype=np.int16)
-    return o.Glyphs().DataAsNumpy().view(np.int16).reshape(DUNGEON_SHAPE)
+        return np.zeros(shape, dtype=dtype)
+    return getattr(o, attr)().DataAsNumpy().view(dtype).reshape(DUNGEON_SHAPE)
```
alexholdenmiller, line 91: `DUNGEON_SHAPE` still in this line

heiner: Ah, many thanks. Fixed!

alexholdenmiller:
excited to try it out!

heiner:
> excited to try it out!

You'll have to approve your requested changes first ;)

heiner:
Nevermind GH just updated.

heiner:
One test failed due to a timeout after 10 mins. I noticed yesterday that test_envs is quite slow, but it wasn't _that_ slow for me. Will re-run test to see if it's just flaky.

heiner:
Haha, the test always ran until done=True (in each combination of archive/no archive & each task). Not good.

PR #58, opened 2020-08-03
Introduce hack to fix a race in the makedef runs.

heiner:
It turns out both makedefs -e and -s create & destroy dat/grep.tmp.
This file creation & deletion can raise. This change hacks it to be
linearized.

This issue mainly shows up when using ninja (>> make).


edran (approved):
LGTM! Thank you for fixing this, I knew this issue existed but didn't quite find the time to look up the right target that was causing the issue.

PR #59, opened 2020-08-04
Libnethack 3.6

heiner:
This is nethack as a library.

Uses:

* fcontext for context switching ("yielding"), originally from Boost.
* dlopen/dlclose to reset global state. This is not POSIX compliant[1], but "works".
  * Even worse, we *copy* the .so file in order for dlopen to create a new global state in a situation when we have several environments in the same process (dlopen looks at the inode or something, sym/hard-linking won't work). That's nuts but appears to work. On Linux, dlmopen and linker namespaces might be an alternative; however I didn't manage to make this work.

## 🚀   Speed measurements indicate a > 10x improvement over the current IPC implementation. 🚀

Speed measurements, raw environment in a Python loop:

```
$ python nle/scripts/play.py -m random -n 3 --no-render -e raw
Episode: 0. Steps: 2139. SPS: 25038.358496
Episode: 1. Steps: 1736. SPS: 29816.063741
Episode: 2. Steps: 1574. SPS: 31407.483889
Finished after 3 episodes. Mean sps: 28753.968709
```

Speed measurements, environment with Gym wrapper in a Python loop:

```
$ python nle/scripts/play.py -m random -n 3 --no-render
Final reward: 0.0
End status: DEATH
Mean reward: -0.006100478468899518
Episode: 0. Steps: 418. SPS: 17717.339515
Final reward: 0.0
End status: DEATH
Mean reward: 0.0023146067415730303
Episode: 1. Steps: 445. SPS: 16878.123638
Final reward: 0.0
End status: DEATH
Mean reward: 0.0019212598425196886
Episode: 2. Steps: 1270. SPS: 14700.374649
Finished after 3 episodes. Mean sps: 16431.945934
```

This implementation currently `cd`s into `HACKDIR`, which changes the `pwd` of the process.

[1]: ["Although a dlclose() operation is not required to remove structures from an address space, neither is an implementation prohibited from doing so."](https://pubs.opengroup.org/onlinepubs/007904975/functions/dlclose.html)


heiner:
To try out:
```shell
python setuy.py install
python nle/tests/test_pynle.py
```

Alternative w/o Python:
```shell
mkdir build
cd build
cmake -GNinja -DPYTHON_EXECUTABLE=`which python` -DPYTHON_LIBRARIES=path/to/libpython3.7m.dylib -DCMAKE_BUILD_TYPE=Debug ..
ninja
./rlmain
```

In other shell:
```shell
cd build
python ../nle/scripts/ttyplay.py -p nle.ttyrec
```

nle/env/base.py
```diff
@@ -294,7 +98,7 @@ class StepStatus(enum.IntEnum):
             "exp_lev",
             "gold",
             "hunger",
-            "killer_name",
+            # "killer_name",
```
alexholdenmiller, line 101: intentionally commented out?

nle/env/base.py
```diff
@@ -348,9 +151,14 @@ def __init__(
...
 
-        self.response = None
+        self.last_observation = None
 
         if archivefile is not None:
+            warnings.warn("Setting archive file not yet implemented")
+            archivefile = None
+
+        if archivefile is not None:
+            warnings.warn("Setting archive file not yet implemented")
```
alexholdenmiller, line 161: repeat code

edran:
This PR:

dragon uh oh

@heiner: is there anything you'd like me to carefully look at right now? Given all the changes, it's really hard to review the PR properly.

nle/__init__.py
```diff
@@ -1,6 +1,6 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 #
 # This is needed to register envs
-import nle.env
+# import nle.env
```
alexholdenmiller, line 4: should just delete if not needed

include/config.h
```diff
@@ -342,7 +350,7 @@
  *      died due to program or system crashes to be resumed from the point
  *      of the last level change, after running a utility program.
  */
-#define INSURANCE /* allow crashed game recovery */
+/* #define INSURANCE allow crashed game recovery */
```
alexholdenmiller, line 353: meant to stay commented?

nle/env/base.py
```diff
@@ -313,21 +117,19 @@ def __init__(
             "chars",
             "colors",
             "specials",
-            "status",
+            "blstats",
             "message",
-            "inventory",
```
alexholdenmiller, line 318: bye bye inventory

nle/env/base.py
```diff
@@ -481,113 +281,118 @@ def step(self, action: int):
...
-        response, done, info = self._perform_known_steps(response, done, info)
+        observation, done = self.env.step(self._actions[action])
+        observation, done = self._perform_known_steps(observation, done)
 
         self._steps += 1
 
-        last_response = self.response
-        self.response = response
+        last_observation = self.last_observation
+        self.last_observation = observation
```
alexholdenmiller, line 290: I do like this rename

nle/env/tasks.py
```diff
@@ -262,32 +250,21 @@ def reset(self, *args, **kwargs):
...
+
+        dungeon_num, dungeon_level = blstats[23:25]
+
+        key = (dungeon_num, dungeon_level)
+        explored = np.sum(glyphs != 0)
+        explored_old = 0
+        if key in self.dungeon_explored:
+            explored_old = self.dungeon_explored[key]
+            reward = explored - explored_old
+            self.dungeon_explored[key] = explored
```
alexholdenmiller, line 268: did you mean to keep these two lines under this conditional block?

heiner:
> This PR (with <3):
> 
> ![dragon uh oh]> 

I knnnnoowwww. Sorry about this! This changes everything, Naomi Klein style. I know this is impossible to review (especially so given that it isn't 100% done yet and many of the code smells still in it are due to that).

But it's like 10-20x faster which is why I was pushing this so hard. And I honestly don't see a way of introducing this (drop zmq, forking, subprocesses and do libraries and dlopen + fcontext fibers instead) in clean small PRs.

> @heiner: is there anything you'd like me to carefully look at right now? Given all the changes, it's really hard to review the PR properly (not that we should do so necessarily now :) ).

One thing that would be cool is if you could clone this and try it out. E.g., produce a new version of the SPS table in the whitepaper? See how you like the files etc regardless of the diffs and where stuff came from.

include/config.h
heiner: Yes, this is related to file locks which we don't want for several reasons.

nle/env/tasks.py
heiner: Good catch, thanks!

nle/env/base.py
heiner: We can re-add the inventory observation once we know what to do with it.

Sadly, this isn't the only "menu" we'll have to deal with: I noticed NetHack will automatically pop up an ingame-window for tiles with lots of objects on them, and there's no way to make it list that with topline messages (although there is https://nethackwiki.com/wiki/Options#pile_limit). So to deal with objects, we will have to think about what to do there.

nle/env/base.py
heiner: It used to be the "response" we got from RPC. Now it's just the observation directly in memory. :)

nle/__init__.py
heiner: Right :)

This was just a leftover-comment from when nothing worked on this branch. Will revert to what it was originally.

win/rl/pynethack.cc
```diff
@@ -0,0 +1,381 @@
...
+                return &def_monsyms[let];
+            },
+            py::return_value_policy::reference)
+        .def_readonly("sym", &class_sym::sym)
+        .def_readonly("name", &class_sym::name)
+        .def_readonly("explain", &class_sym::explain)
+        .def("__repr__", [](const class_sym &cs) {
+            return "<nethack.pynle.class_sym sym='" + std::string(1, cs.sym)
+                   + "' explain='" + std::string(cs.explain) + "'>";
+        });
```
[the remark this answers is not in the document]
heiner: Yes I renamed helper.cc and added the NLE class to it -- in libnethack, this is how we interact with NetHack, so this code is no longer just a "helper" module defining a few constants.

nle/env/base.py
```diff
@@ -481,147 +268,159 @@ def step(self, action: int):
...
+            # This fails if the agent picks up a scroll of scare
+            # monster at the 0th turn and gets asked to name it.
+            # Hence the defensive iteration above.
+            self.last_observation, done = self.env.step(ASCII_SPACE)
             assert not done, "Game ended unexpectedly"
+        else:
+            warnings.warn(
+                "Not in moveloop after 1000 tries, aborting (ttyrec: %s)." % new_ttyrec
+            )
+            return reset()
```
alexholdenmiller, line 380: I suspect this is the cause of hackrl/issues/170

heiner: Thanks! Fixed in 5260b13.

nle/env/base.py
rockt: I think function would suffice.

nle/env/base.py
```diff
@@ -481,147 +268,160 @@ def step(self, action: int):
...
 
         self._steps = 0
-        self._info = self.env._info.copy()
 
-        self._info["seeds"] = {}
-        for k in nethack.SEED_KEYS:
-            self._info["seeds"][k] = getattr(self.response.Seeds(), k.capitalize())()
-
-        while not _in_moveloop(self.response):
+        for _ in range(1000):
```
rockt, line 366: What's the 1000?

nle/env/base.py
```diff
@@ -631,23 +431,37 @@ def render(self, mode="human"):
                        "ansi", otherwise a ``ValueError`` is raised.
 
         """
+        chars_index = self._observation_keys.index("chars")
         if mode == "human":
-            nhprint.print_message(self.response)
+            message_index = self._observation_keys.index("message")
```
rockt, line 436: Default color rendering? Nice!

nle/env/tasks.py
```diff
@@ -200,27 +181,33 @@ class NetHackGold(NetHackScore):
...
         """Difference between previous gold and new gold."""
         del end_status  # Unused
+        if not self.env.in_normal_game():
+            # Before game started or after it ended stats are zero.
+            return 0.0
+
+        old_blstats = last_observation[self._blstats_index]
+        blstats = observation[self._blstats_index]
+
+        old_gold = old_blstats[14]
```
rockt, line 207: I wonder whether we should turn this into methods instead of indexing using a hard-coded number.

win/rl/winrl.cc
```diff
@@ -195,294 +206,175 @@ class NetHackRL
...
+            u.uac,               /* armor_class       */
+            Upolyd ? (int) mons[u.umonnum].mlevel : 0, /* monster_level     */
+            u.ulevel,                                  /* experience_level  */
+            u.uexp,                                    /* experience_points */
+            moves,                                     /* time              */
+            u.uhs,                                     /* hunger state      */
+            near_capacity(),                           /* carrying_capacity */
+            u.uz.dnum,
+            u.uz.dlevel,
+        };
```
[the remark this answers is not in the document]
rockt: Yeah, we should switch to dnum + dlevel + x + y for RIDE.

rockt (approved):
Impressive work @heiner!

nle/env/base.py
heiner: As the comment says:

The purpose of this loop is to get out of early-game situations. However:

```python
            # This fails if the agent picks up a scroll of scare
            # monster at the 0th turn and gets asked to name it.
            # Hence the defensive iteration above.
            # TODO: Detect this 'in_getlin' situation and handle it.
```

So as a defensive measure, we only try 1000 times, then throw away the offending restart.

nle/env/tasks.py
heiner: Yes, except that it's measurably faster to call fewer functions ...

rockt: How about we give this a name though and define a constant for each of the indices, or would this slow down things too?

heiner: No that would be fine.

My excuse has been shown as phony. :D

heiner: (But more long term I think it might be a good idea to move the tasks into C++ once they have stabilized.)

PR #60, opened 2020-08-14
Fix and test permonst deallocation issue

heiner:


nle/tests/test_nethack.py
```diff
@@ -89,6 +89,21 @@ def test_permonst(self):
...
+            self.assertEqual(mon.mname, "giant ant")
+
+    def test_illegal_numbers(self):
+        with self.assertRaisesRegex(IndexError, "NUMMONS"):
+            nethack.permonst(10000)
+
+        with self.assertRaisesRegex(IndexError, "NUMMONS"):
+            nethack.permonst(-1)
+
+        with self.assertRaisesRegex(IndexError, "MAXMCLASSES"):
```
edran, line 104: Just to be consistent, I would probably add an exception check for negative values for `from_mlet`.

nle/tests/test_nethack.py
```diff
@@ -89,6 +89,21 @@ def test_permonst(self):
...
         self.assertEqual(mon.mname, "killer bee")
 
+        # Test https://github.com/pybind/pybind11/issues/2394
+        for _ in range(2):
+            mon = nethack.permonst(0)
+            self.assertEqual(mon.mname, "giant ant")
+
+    def test_illegal_numbers(self):
+        with self.assertRaisesRegex(IndexError, "NUMMONS"):
+            nethack.permonst(10000)
```
edran, line 99: Do (or should?) we have access to `NUMMONS` at runtime here? That way we could check for `upper_bound + 1`.

EDIT: better question

edran (approved):
LGTM! Thanks, well spotted!

win/rl/helper.cc
```diff
@@ -124,9 +124,18 @@ PYBIND11_MODULE(helper, m)
...
-    py::class_<permonst, std::unique_ptr<permonst, py::nodelete> >(m,
-                                                                   "permonst")
-        .def(py::init([](int index) -> permonst * { return &mons[index]; }))
+    py::class_<permonst>(m, "permonst")
+        .def(
+            "__init__",
+            [](py::detail::value_and_holder &v_h, int index) {
+                if (index < 0 || index >= NUMMONS)
+                    throw std::out_of_range(
+                        "Index should be between 0 and NUMMONS");
```
alexholdenmiller, line 133: should the error include what NUMMONS is, e.g. "Index should be between 0 and NUMMONS (int)"

win/rl/helper.cc
```diff
@@ -159,7 +168,12 @@ PYBIND11_MODULE(helper, m)
     py::class_<class_sym>(m, "class_sym")
         .def_static(
             "from_mlet",
-            [](char let) -> const class_sym * { return &def_monsyms[let]; },
+            [](char let) -> const class_sym * {
+                if (let < 0 || let >= MAXMCLASSES)
+                    throw std::out_of_range(
+                        "Argument should be between 0 and MAXMCLASSES");
```
alexholdenmiller, line 174: same as above

nle/tests/test_nethack.py
heiner: I would if I could -- but mlet is a char which in Python gets mapped to a string of length 1. I cannot produce negative values there :D

nle/tests/test_nethack.py
heiner: Done.

win/rl/helper.cc
heiner: Kinda weird but okay.

heiner: Ah I begin to understand what you meant. Fixed again.

PR #62, opened 2020-08-26
Very important fix: nhdat did not contain all files necessary.

heiner:
**Previously, we had a too-small `nhdat` file and the game would crash or otherwise get into a bad state as soon as we entered e.g. the Gnomish mines.**

This has been the case since we switched to cmake, but it wasn't easy to detect as we need agents that get as far first.

This fixes the cmake logic to produce a `nhdat` with all required files.

**However**, I don't necessarily like that we build everything in the source directory. We should fix dat/CMakeLists.txt to copy the source dat/ directory into the build directory and run all commands there. I suspect this may make the logic in this file easier.

### DIFF
--- a/dat/CMakeLists.txt
+++ b/dat/CMakeLists.txt
@@ -1,4 +1,10 @@
+# TODO: This creates lots of files in ${NLE_DAT}, i.e. in the source dir. We
+# should create files in the binary dir only. NetHack's utils require their
+# arguments to be in the same local directory, so this requires copying files
+# from ${NLE_DAT} to ${NLE_DAT_GEN}
+
 set(VARDATND) # NOTE: this is empty for us
+
 set(VARDATD
     bogusmon
     data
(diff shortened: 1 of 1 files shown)


rockt (approved):

dat/CMakeLists.txt
```diff
@@ -1,4 +1,10 @@
+# TODO: This creates lots of files in ${NLE_DAT}, i.e. in the source dir. We
+# should create files in the binary dir only. NetHack's utils require their
+# arguments to be in the same local directory, so this requires copying files
+# from ${NLE_DAT} to ${NLE_DAT_GEN}
```
edran, line 4: [draft comment -- putting this down as I read the rest of the PR]
This comment confuses me. The way I setup the build process is that there are temporary steps where the nethack src dirs are dirty (because of the various generated data files and headers), but they _should_ get removed or moved into the build dir as soon as it is possible.

This is not ideal, however it's not easy to fix without modifying makedefs, since the binary hardcodes paths wrt. the source dir location, not `$cwd` (it is however fairly trivial to fix it if we are OK with modifying `makedefs` and such).

edran: I see, with this PR `*.lev` files are generated in `dat/`.

edran (approved):
>However, I don't necessarily like that we build everything in the source directory.

Setting this aside (given #63), LGTM.

PR #64, opened 2020-09-18
Forward wizard mode through nle, add tests.

yobibyte:


nle/tests/test_envs.py
```diff
@@ -75,25 +75,26 @@ def compare_rollouts(env0, env1, max_rollout_len):
 
 
 @pytest.mark.parametrize("env_name", get_nethack_env_ids())
+@pytest.mark.parametrize("wizard", [False, True])
```
heiner, line 78: Thanks for the elaborate test!

I'm a little bit reluctant though to double the runtime of the test like this, although obviously it gives optimal test coverage this way.  Perhaps we could not add `wizard` to `TestGymEnvRollout`?

heiner (approved):
Thanks!

Small question about the test below.

nle/tests/test_envs.py
yobibyte: I was not sure what wizard-mode affects under the hood, that's why I added the option. I can remove not to increase the testing time.

rockt (approved):

heiner (approved):
Thanks! Looks good.

PR #70, opened 2020-09-21
Re-add inventory: Add inv_glyphs observation.

heiner:
Stack from [ghstack](https://github.com/ezyang/ghstack):
* #77 Bump version to 0.5.1.
* #76 Update lint_python workflow to print diff when black failed.
* #75 Add inv_* observations to base.py.
* #74 Handle keep_alive logic explicitly.
* #73 Add inv_strs observation.
* #72 Use update_inventory for inventory observations.
* #71 Add inv_letters and inv_oclasses observations.
* **#70 Re-add inventory: Add inv_glyphs observation.**

PR #71, opened 2020-09-21
Add inv_letters and inv_oclasses observations.

heiner:
* #77 Bump version to 0.5.1.
* #75 Add inv_* observations to base.py.
* #74 Handle keep_alive logic explicitly.
* **#71 Add inv_letters and inv_oclasses observations.**

### DIFF
--- a/include/nleobs.h
+++ b/include/nleobs.h
@@ -10,16 +10,18 @@
 typedef struct nle_observation {
     int action;
     int done;
-    char in_normal_game;     /* Bool indicating if other obs are set. */
-    short *glyphs;           /* Size ROWNO * (COLNO - 1) */
-    unsigned char *chars;    /* Size ROWNO * (COLNO - 1) */
-    unsigned char *colors;   /* Size ROWNO * (COLNO - 1) */
-    unsigned char *specials; /* Size ROWNO * (COLNO - 1) */
-    unsigned char *message;  /* Size 256 */
-    long *blstats;           /* NLE_BLSTATS_SIZE */
-    int *program_state;      /* NLE_PROGRAM_STATE_SIZE */
-    int *internal;           /* NLE_INTERNAL_SIZE */
-    short *inv_glyphs;       /* NLE_INVENTORY_SIZE */
+    char in_normal_game;         /* Bool indicating if other obs are set. */
+    short *glyphs;               /* Size ROWNO * (COLNO - 1) */
+    unsigned char *chars;        /* Size ROWNO * (COLNO - 1) */
+    unsigned char *colors;       /* Size ROWNO * (COLNO - 1) */
+    unsigned char *specials;     /* Size ROWNO * (COLNO - 1) */
+    unsigned char *message;      /* Size 256 */
+    long *blstats;               /* NLE_BLSTATS_SIZE */
+    int *program_state;          /* NLE_PROGRAM_STATE_SIZE */
+    int *internal;               /* NLE_INTERNAL_SIZE */
+    short *inv_glyphs;           /* NLE_INVENTORY_SIZE */
+    unsigned char *inv_letters;  /* NLE_INVENTORY_SIZE */
+    unsigned char *inv_oclasses; /* NLE_INVENTORY_SIZE */
 } nle_obs;
 
 typedef struct {
--- a/nle/nethack/nethack.py
+++ b/nle/nethack/nethack.py
@@ -29,6 +29,8 @@ OBSERVATION_DESC = {
     "program_state": dict(shape=PROGRAM_STATE_SHAPE, dtype=np.int32),
     "internal": dict(shape=INTERNAL_SHAPE, dtype=np.int32),
     "inv_glyphs": dict(shape=INV_GLYPHS_SHAPE, dtype=np.int16),
+    "inv_letters": dict(shape=INV_GLYPHS_SHAPE, dtype=np.uint8),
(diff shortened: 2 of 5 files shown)


include/nleobs.h
```diff
@@ -10,16 +10,18 @@
...
-    char in_normal_game;     /* Bool indicating if other obs are set. */
-    short *glyphs;           /* Size ROWNO * (COLNO - 1) */
-    unsigned char *chars;    /* Size ROWNO * (COLNO - 1) */
-    unsigned char *colors;   /* Size ROWNO * (COLNO - 1) */
-    unsigned char *specials; /* Size ROWNO * (COLNO - 1) */
-    unsigned char *message;  /* Size 256 */
-    long *blstats;           /* NLE_BLSTATS_SIZE */
-    int *program_state;      /* NLE_PROGRAM_STATE_SIZE */
-    int *internal;           /* NLE_INTERNAL_SIZE */
-    short *inv_glyphs;       /* NLE_INVENTORY_SIZE */
```
yobibyte, line 22: That's weird. Why does the diff include these lines?

nle/nethack/nethack.py
```diff
@@ -29,6 +29,8 @@
     "program_state": dict(shape=PROGRAM_STATE_SHAPE, dtype=np.int32),
     "internal": dict(shape=INTERNAL_SHAPE, dtype=np.int32),
     "inv_glyphs": dict(shape=INV_GLYPHS_SHAPE, dtype=np.int16),
+    "inv_letters": dict(shape=INV_GLYPHS_SHAPE, dtype=np.uint8),
+    "inv_oclasses": dict(shape=INV_GLYPHS_SHAPE, dtype=np.uint8),
```
yobibyte, line 33: Copypaste typo for shape?

win/rl/pynethack.cc
```diff
@@ -222,11 +227,16 @@ PYBIND11_MODULE(_pynethack, m)
...
-             py::arg("inv_glyphs") = py::none(), py::keep_alive<1, 2>(),
-             py::keep_alive<1, 3>(), py::keep_alive<1, 4>(),
-             py::keep_alive<1, 5>(), py::keep_alive<1, 6>(),
-             py::keep_alive<1, 7>(), py::keep_alive<1, 8>(),
-             py::keep_alive<1, 9>(), py::keep_alive<1, 10>())
+             py::arg("inv_glyphs") = py::none(),
+             py::arg("inv_letters") = py::none(),
+             py::arg("inv_oclasses") = py::none(),
+
+             py::keep_alive<1, 2>(), py::keep_alive<1, 3>(),
```
yobibyte, line 234: Can you, please, explain what keep_alive does?

win/rl/winrl.cc
```diff
@@ -367,6 +367,27 @@ NetHackRL::fill_obs(nle_obs *obs)
...
+    if (obs->inv_letters) {
+        /* This iterates over the invent linked list once per inv observation
+         instead of only once. Could instead also check all the observations
+         and only copy if they all exist.*/
+        int i = 0;
+        for (const struct obj *otmp = invent; otmp; otmp = otmp->nobj) {
+            obs->inv_letters[i++] = otmp->invlet;
+        }
+        for (; i < NLE_INVENTORY_SIZE; ++i) {
+            obs->inv_letters[i] = 0;
```
yobibyte, line 379: Is it a stub for future changes? Why is it zero?

yobibyte (changes requested):
Thanks! Can you, please, check the copypaste of shape constants in nethack.py?

include/nleobs.h
heiner: The reason is the changed space between the code and the comments. The reason of _that_ is that clang-format decided to change this space for whatever reason.

nle/nethack/nethack.py
heiner: Good point. It's literally the same shape but we may want to make that more or less obvious.

Once we have the strings as well their shape will be different.

win/rl/pynethack.cc
heiner: See https://pybind11.readthedocs.io/en/stable/advanced/functions.html#keep-alive. Basically, we need to keep track of the lifetimes of objects in Python and C++. Python's objects are refcounted, and this instruction (I believe), adds a refcount to the argument (the numpy arrays we fil things in) that will be decref'd when the `Nethack` object gets destroyed[1]. That's not 100% the right behaviour when `set_buffers` is called several times but seemed easy enough.

[1] Yes, this underlying logic fills in a supplied buffer w/o additional copies for additional performance. That can be dangerous when the user stores return values longer than one step w/o copying, which is why we also have a `copy=True` option.

win/rl/winrl.cc
heiner: This is the padding -- the observation has a length of NLE_INVENTORY_SIZE (which I chose to be 55 which is > 2 * 26 for each letter in the alphabet plus its upper case version, plus some margin of error). The first entries get filled via the linked list with current-item pointer `otmp`. Whatever remains gets set to `0`.

yobibyte: Oh yeah, I did not pay attention to i++ above. Thanks!

nle/nethack/nethack.py
heiner: (Fixed in https://github.com/facebookresearch/nle/pull/73).

PR #73, opened 2020-09-22
Add inv_strs observation.

heiner:
Stack from [ghstack](https://github.com/ezyang/ghstack):
* #77 Bump version to 0.5.1.
* #76 Update lint_python workflow to print diff when black failed.
* #75 Add inv_* observations to base.py.
* #74 Handle keep_alive logic explicitly.
* **#73 Add inv_strs observation.**
* #72 Use update_inventory for inventory observations.
* #71 Add inv_letters and inv_oclasses observations.
* #70 Re-add inventory: Add inv_glyphs observation.

PR #74, opened 2020-09-22
Handle keep_alive logic explicitly.

heiner:
* #77 Bump version to 0.5.1.
* #75 Add inv_* observations to base.py.
* **#74 Handle keep_alive logic explicitly.**
* #71 Add inv_letters and inv_oclasses observations.

### DIFF
--- a/win/rl/pynethack.cc
+++ b/win/rl/pynethack.cc
@@ -33,7 +33,7 @@ checked_conversion(py::object obj, const std::vector<ssize_t> &shape)
 {
     if (obj.is_none())
         return nullptr;
-    auto array = py::array::ensure(obj.release());
+    py::array array = py::array::ensure(obj.release());
     if (!array)
         throw std::runtime_error("Numpy array required");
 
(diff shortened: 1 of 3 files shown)


win/rl/pynethack.cc
```diff
@@ -33,7 +33,7 @@ checked_conversion(py::object obj, const std::vector<ssize_t> &shape)
 {
     if (obj.is_none())
         return nullptr;
-    auto array = py::array::ensure(obj.release());
+    py::array array = py::array::ensure(obj.release());
```
yobibyte, line 36: Can you explain why this way is better?

win/rl/pynethack.cc
```diff
@@ -117,28 +117,32 @@ class Nethack
...
-        obs_.glyphs = checked_conversion<int16_t>(std::move(glyphs), dungeon);
-        obs_.chars = checked_conversion<uint8_t>(std::move(chars), dungeon);
-        obs_.colors = checked_conversion<uint8_t>(std::move(colors), dungeon);
-        obs_.specials =
-            checked_conversion<uint8_t>(std::move(specials), dungeon);
-        obs_.blstats = checked_conversion<long>(std::move(blstats),
-                                                { NLE_BLSTATS_SIZE });
-        obs_.message =
-            checked_conversion<uint8_t>(std::move(message), { 256 });
+        obs_.glyphs = checked_conversion<int16_t>(glyphs, dungeon);
```
yobibyte, line 120: Why do we need to remove std::move?

yobibyte (approved):

win/rl/pynethack.cc
heiner: I had to look up what `py::array::ensure` actually does and it turns out it returns a `py::array`, which is only 5 more characters than `auto` and would have prevented me wondering what it returns. Hence I changed it.

win/rl/pynethack.cc
heiner: Good question.

A moved-from object in C++ becomes "empty", hence the `py_buffers_` assignment further down wouldn't be sensible. By removing the `std::move` we copy the `py::object`s, which increases their refcounts. Strictly speaking that's not necessary here, but this isn't performance-relevant code

PR #75, opened 2020-09-22
Add inv_* observations to base.py.

heiner:
* #77 Bump version to 0.5.1.
* **#75 Add inv_* observations to base.py.**
* #74 Handle keep_alive logic explicitly.
* #71 Add inv_letters and inv_oclasses observations.

We are now using a flat instead of a nested version for simplicity.

### DIFF
--- a/nle/tests/test_envs.py
+++ b/nle/tests/test_envs.py
@@ -119,6 +119,37 @@ class TestGymEnv:
             assert "playmode:debug" not in env.env._options
 
 
+@pytest.mark.parametrize("env_name", [e for e in get_nethack_env_ids() if "Score" in e])
+class TestBasicGymEnv:
+    def test_inventory(self, env_name):
+        env = gym.make(
+            env_name,
+            observation_keys=(
+                "chars",
+                "inv_glyphs",
+                "inv_strs",
+                "inv_letters",
+                "inv_oclasses",
+            ),
+        )
+        obs = env.reset()
+
+        found = dict(spellbook=0, apple=0)
+        for line in obs["inv_strs"]:
+            if np.all(line == 0):
+                break
+            for key in found:
+                if key in line.tobytes().decode("utf-8"):
+                    found[key] += 1
+
+        for key, count in found.items():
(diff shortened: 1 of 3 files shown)


nle/tests/test_envs.py
```diff
@@ -119,6 +119,37 @@ def test_default_wizard_mode(self, env_name, wizard):
...
+            for key in found:
+                if key in line.tobytes().decode("utf-8"):
+                    found[key] += 1
+
+        for key, count in found.items():
+            assert key == key and count > 0
+
+        assert "inv_strs" in obs
+        assert obs["inv_letters"][0] == ord("a")
+        assert obs["inv_oclasses"][0] == nethack.ARMOR_CLASS
```
yobibyte, line 150: Does every agent has this in the inventory?

nle/tests/test_envs.py
```diff
@@ -119,6 +119,37 @@ def test_default_wizard_mode(self, env_name, wizard):
...
+        found = dict(spellbook=0, apple=0)
+        for line in obs["inv_strs"]:
+            if np.all(line == 0):
+                break
+            for key in found:
+                if key in line.tobytes().decode("utf-8"):
+                    found[key] += 1
+
+        for key, count in found.items():
+            assert key == key and count > 0
```
yobibyte, line 146: Why do we expect agent to have an apple?

yobibyte (approved):
Thanks!

nle/tests/test_envs.py
heiner: Our default character class in NLE is a monk, mainly because the monk does martial arts and therefore doesn't need a weapon to create large damage. As per the wiki, https://nethackwiki.com/wiki/Monk, monks start with gloves and a robe. The inventory also starts with `a`, and worn stuff comes first. Hence, the first entry will be that.

nle/tests/test_envs.py
heiner: This test implicitly assumes we get a monk. Monks always start with apples. :)

PR #81, opened 2020-09-23
Add objclass and objdescr as Python classes.

heiner:
This is a bit more complicated than permonst as the user can
"name" and "discover" an object and hence change it. This here does
not interact with any real running NetHack, so these fields won't
change and/or have any real information.

### DIFF
--- a/nle/tests/test_nethack.py
+++ b/nle/tests/test_nethack.py
@@ -247,6 +247,15 @@ class TestNethackSomeObs:
         assert internal[3] == 1  # xwaitforspace
 
 
+def get_object(name):
+    for index in range(nethack.NUM_OBJECTS):
+        obj = nethack.objclass(index)
+        if nethack.OBJ_NAME(obj) == name:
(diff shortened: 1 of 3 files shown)


nle/tests/test_nethack.py
```diff
@@ -296,3 +305,29 @@ def test_illegal_numbers(self):
...
+        od = nethack.objdescr.from_idx(0)
+        assert od.oc_name == "strange object"
+        assert od.oc_descr is None
+
+        elven_dagger = get_object("elven dagger")
+        od = nethack.objdescr.from_idx(elven_dagger.oc_name_idx)
+        assert od.oc_name == "elven dagger"
+        assert od.oc_descr == "runed dagger"
+
+        # Example of how to do this with glyphs.
```
yobibyte, line 329: Thanks for this example!

yobibyte (approved):
Looks good! Thanks!

PR #83, opened 2020-09-28
Add RUSH action.

yobibyte:


heiner:
Should we add `m`, `M`, and `F` too?

https://github.com/facebookresearch/nle/blob/master/src/cmd.c#L4296

heiner (approved):
LGTM.

PR #85, opened 2020-09-30
Update README, docs, GitHub workflows, Dockerfiles.

heiner:
Remove all mentions of `zmq`/`zeromq` and `flatbuffers` as we no longer depend on those libraries.

### DIFF
--- a/.github/workflows/test_package.yml
+++ b/.github/workflows/test_package.yml
@@ -20,6 +20,9 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Ensure latest pip & wheel
       run: "python -m pip install -q --upgrade pip wheel"
+    - name: Install dependencies
+      run: |
+        brew install cmake
     - name: Clone repo
       uses: actions/checkout@v2
     - name: Install nle via pip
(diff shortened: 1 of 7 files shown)


.github/workflows/test_package.yml
```diff
@@ -22,12 +22,10 @@ jobs:
...
-          /usr/local/include
+        brew install cmake
     - name: Clone repo
       uses: actions/checkout@v2
-    - name: Install from repo in test mode
+    - name: Install nle via pip
       run: "pip install nle"
     - name: Check nethack is installed
-      run: "type -p nethack"
+      run: "python -c 'import nle; import gym; gym.make(\"NetHack-v0\")'"
```
rockt, line 31: Is there any test associated with this or do we simply expect it not to crash?

rockt (approved):
Not much to comment on. Looks good.

.github/workflows/test_package.yml
heiner: This is the test for the package and we just hope it doesn't crash but instead is able to load the environment.

PR #88, opened 2020-10-06
Add PRAY to autoskip exceptions.

yobibyte:


nle/env/base.py
```diff
@@ -214,7 +220,9 @@ def __init__(
 
         space_dict = {
             "glyphs": gym.spaces.Box(
-                low=0, high=nethack.MAX_GLYPH, **nethack.OBSERVATION_DESC["glyphs"]
+                low=0,
+                high=nethack.MAX_GLYPH,
+                **nethack.OBSERVATION_DESC["glyphs"],
```
heiner, line 225: Wait, did black do this?

There are versions of Python that really really dislike the comma after `**kwargs`.

nle/env/base.py
```diff
@@ -539,9 +551,16 @@ def _perform_known_steps(self, observation, done, exceptions=True):
...
                     msg = bytes(observation[self._message_index])
-                    # Allow agent to select stuff to eat, attack, and to
-                    # select directions.
-                    if b"eat" in msg or b"attack" in msg or b"direction?" in msg:
+                    # Do not skip some questions to allow agent to select
+                    # stuff to eat, attack, and to select directions.
+                    decline = True
+                    # I would go with this to avoid surprises when we allow all
+                    # but restrict some.
+                    for el in [b"eat", b"attack", b"direction?", b"pray"]:
```
heiner, line 559: I find this a bit complicated.

We could do something along the lines of `any(el in s for el in ("y", "x", "a"))`?

Also does this do what you want? If `exceptions` is false, we still decline?
Would make sense to refactor this into

```
if not exceptions:
  observation, done = self.env.step(ASCII_ESC)
  continue

# exceptions is True here, so do something else.
```

heiner (commented):
Agree in principle, would like slight refactoring for my brain to be able to understand this though.

(`_perform_known_steps` has in the past been a source of irritating trouble as it's very easy to get wrong but hard to read the logic.)

nle/env/base.py
yobibyte: I think, black did this when I run it with '-l 80'. Right now black uses the default option for line width.

heiner: But the comma after the `**kwargs` arg?

See https://github.com/psf/black/issues/419 -- what version of `black` are you using? I am not sure we should have commas there, and at any rate there's no need for this PR to change these lines.

nle/env/base.py
yobibyte: That's a good catch! I fixed the logic. Now when calling step() we either allow all the yn questions or skip those that are not in the exceptions list.

heiner (approved):
LGTM.

PR #94, opened 2020-10-19
New ttyplay features

heiner:
This change makes `l` in `nle-ttyplay` work again (for skipping forward).

It also supports playing bzip2'd and gzip'ed input files.

### DIFF
--- a/nle/scripts/ttyplay.py
+++ b/nle/scripts/ttyplay.py
@@ -3,6 +3,7 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 import argparse
 import os
+import re
 import select
 import struct
 import termios
@@ -97,7 +98,7 @@ def read_header(fd, peek=False, no_input=False):
         yield timestamp, length, channel
 
 
-CLRCODE = b"\033[2J"
+CLRCODE = re.compile(rb"\033\[2?J")  # https://stackoverflow.com/a/37778152/1136208
 
 
 def process(fd):
@@ -128,8 +129,7 @@ def process(fd):
         if jump == 0 and prev is not None:
             speed, drift, jump = wait(timestamp - prev, speed, drift)
 
-        offset = data.find(CLRCODE)
-        if offset >= 0:
+        if CLRCODE.search(data):
             clrscreen.append((lastpos, prev))
             if jump > 0:
                 jump = 0
--- a/nle/scripts/ttyplay.py
+++ b/nle/scripts/ttyplay.py
@@ -2,6 +2,8 @@
 #
 # Copyright (c) Facebook, Inc. and its affiliates.
 import argparse
+import bz2
+import gzip
 import os
 import re
 import select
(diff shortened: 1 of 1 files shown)


nle/scripts/ttyplay.py
```diff
@@ -97,10 +100,10 @@ def read_header(fd, peek=False, no_input=False):
         yield timestamp, length, channel
 
 
-CLRCODE = b"\033[2J"
+CLRCODE = re.compile(rb"\033\[2?J")  # https://stackoverflow.com/a/37778152/1136208
```
yobibyte, line 103: Why do we expect multiple 2s here? (I'm speaking about the question mark).

yobibyte (approved):

nle/scripts/ttyplay.py
condnsdmatters: maybe ive misunderstood but the question mark makes the 2 optional (1 or 0) as per the stack overflow

nle/scripts/ttyplay.py
```diff
@@ -165,10 +167,14 @@ def main():
     FLAGS = parser.parse_args()
 
     if FLAGS.filename == "-":
-        fd = os.dup(0)
+        f = os.fdopen(os.dup(0), "rb")
```
condnsdmatters, line 170: just curious as two what youre doing here?  is this is a command line interface that allows you to read from std in? if so, why are you mapping std out to std in below? appreciate its not a real change, just wondering.

condnsdmatters (approved):
LGTM, though I havent tested it runs.

nle/scripts/ttyplay.py
heiner: The question mark makes the `2` optional as @condnsdmatters says. As per the link and the commit's message (https://github.com/facebookresearch/nle/pull/94/commits/2e3526777971bb5efe856a87ec00a4840fb16dba), this detects more _clear screen_ events: For instance, in the ttyrec you sent, @yobibyte, I saw `rb"\x1b[H\x1b[J"` as a clear screen event that we didn't detect previously. Now we can jump around in this kind of ttyrec as well (we can only ever jump to clear screen events because only at those points do we know the state of the "canvas").

nle/scripts/ttyplay.py
heiner: That's a good question. This is some POSIX magic here: Doing it this way allows us to run this via

```shell
python nle/scripts/ttyplay.py - < my-recording.ttyrec
```

and still do the `ioctl` on fd 0 and *use the keyboard to jump around*. Those keyboard actions are also read from fd 0. I am honestly not quite sure why this works. I think I took the idea from here: https://github.com/ovh/ovh-ttyrec/blob/master/ttyplay.c#L306

nle/scripts/ttyplay.py
yobibyte: Sorry, my bad. I understand it now.

PR #97, opened 2020-10-21
Fix terminal size at 80x24, write bzip2'ed ttyrecs

heiner:
* Setting the terminal size makes sure our ttyrecs will be 80x24, which is both the minimum size and the size of the majority of ttyrecs on alt.org
* Writing bzip2'd ttyrecs should save up to 80% of disk space. Only downside is that we can no longer peek into running ttyrecs this way, which was sometimes useful for debugging. To go back to that mode, we can `#undefine NLE_BZ2_TTYRECS`.

### DIFF
--- a/include/nle.h
+++ b/include/nle.h
@@ -1,6 +1,8 @@
 #ifndef NLE_H
 #define NLE_H
 
+#define NLE_BZ2_TTYRECS
+
 #include <stdio.h>
 
 #include <fcontext/fcontext.h>
@@ -20,6 +22,10 @@ typedef struct nle_globals {
     char outbuf[BUFSIZ];
     char *outbuf_write_ptr;
     char *outbuf_write_end;
+
+#ifdef NLE_BZ2_TTYRECS
+    void *ttyrec_bz2;
+#endif
 
     boolean done;
     nle_obs *observation;
--- a/include/unixconf.h
+++ b/include/unixconf.h
@@ -139,7 +139,7 @@
 #define TIMED_DELAY
 #endif
 
-/* #define AVOID_WIN_IOCTL */ /* ensure USE_WIN_IOCTL remains undefined */
+#define AVOID_WIN_IOCTL /* ensure USE_WIN_IOCTL remains undefined */
 
 /*
  * If you define MAIL, then the player will be notified of new mail
--- a/nle/env/base.py
+++ b/nle/env/base.py
@@ -201,7 +201,7 @@ class NLE(gym.Env):
 
         if self.savedir:
             self._ttyrec_pattern = os.path.join(
-                self.savedir, "nle.%i.%%i.ttyrec" % os.getpid()
+                self.savedir, "nle.%i.%%i.ttyrec.bz2" % os.getpid()
             )
             ttyrec = self._ttyrec_pattern % 0
         else:
--- a/src/nle.c
+++ b/src/nle.c
@@ -1,8 +1,7 @@
 
 #include <assert.h>
-#include <sys/time.h>
-
 #include <string.h>
+#include <sys/time.h>
 
 #define NEED_VARARGS
 #include "hack.h"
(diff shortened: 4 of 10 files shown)


include/unixconf.h
```diff
@@ -139,7 +139,7 @@
 #define TIMED_DELAY
 #endif
 
-/* #define AVOID_WIN_IOCTL */ /* ensure USE_WIN_IOCTL remains undefined */
+#define AVOID_WIN_IOCTL /* ensure USE_WIN_IOCTL remains undefined */
```
condnsdmatters, line 142: This comment could probably be added to, to help future clarity.
```suggestion
/* 
 * ensure USE_WIN_IOCTL remains undefined
 * USE_WIN_IOCTL can set LI and CO global variables for terminal size, which we fix in nle.c
 */
#define AVOID_WIN_IOCTL 
```

include/nle.h
```diff
@@ -21,6 +23,10 @@ typedef struct nle_globals {
     char *outbuf_write_ptr;
     char *outbuf_write_end;
 
+#ifdef NLE_BZ2_TTYRECS
+    void *ttyrec_bz2;
+#endif
+
```
condnsdmatters, line 29:
```suggestion
#ifdef NLE_BZ2_TTYRECS
    void *ttyrec_bz2;
#else 
    FILE *ttyrec;
#endif

```

I'm curious: why aren't we doing ^^^ ? We never write to the file if the macro is set. Do we not just run the risk of confusion by keeping this pointer around?

src/nle.c
```diff
@@ -31,6 +34,12 @@ init_nle(FILE *ttyrec)
     assert(ttyrec != NULL);
     nle->ttyrec = ttyrec;
```
condnsdmatters, line 35: same point as above here.

condnsdmatters (approved):
Generally looks good to me (the odd nit) but I wish there was a test. :(. A bit of a pain to do, I will try to run by hand to verify.

include/unixconf.h
condnsdmatters: formatting of the comment can obvs change.

nle/env/base.py
```diff
@@ -201,7 +201,7 @@ def __init__(
 
         if self.savedir:
             self._ttyrec_pattern = os.path.join(
-                self.savedir, "nle.%i.%%i.ttyrec" % os.getpid()
+                self.savedir, "nle.%i.%%i.ttyrec.bz2" % os.getpid()
```
rockt, line 204: pid is always going to be unique within the folder? Are we saving all ttyrecs to there or only the last N?

src/nle.c
```diff
@@ -31,6 +34,12 @@ init_nle(FILE *ttyrec)
...
 
+#ifdef NLE_BZ2_TTYRECS
+    int bzerror;
+    nle->ttyrec_bz2 = BZ2_bzWriteOpen(&bzerror, ttyrec, 9, 0, 0);
```
rockt, line 39: What's the 9?

rockt (approved):

include/unixconf.h
heiner: I see where you're coming from, but I'd like to stick with the policy of not unduly changing the original NetHack source files, and this is one of them.

include/nle.h
heiner: Good question. Who exactly is responsible for this `FILE` is a bit of a tricky question. For now, I'd like to keep it around for debugging and to allow flushing (although we don't do that currently).

nle/env/base.py
heiner: Currently we are either saving all ttyrecs or none.

The pid is going to be the process id of the user's process which won't be reassigned until that process stops.

src/nle.c
heiner: The block size. 9 is the highest and also what alt.org uses.

http://www.cs.cmu.edu/afs/cs/project/pscico-guyb/realworld/99/code/bzip2-0.9.5c/manual_3.html#SEC30

PR #100, opened 2020-10-23
Add inv letter to rendering.

yobibyte:
Useful for testing when selecting items in the inventory. Have no idea how to test this programmatically, but I did test this with play.py.

### DIFF
--- a/nle/env/base.py
+++ b/nle/env/base.py
@@ -477,12 +477,17 @@ class NLE(gym.Env):
             print(message[: message.index(b"\0")])
             try:
                 inv_strs_index = self._observation_keys.index("inv_strs")
+                inv_letters_index = self._observation_keys.index("inv_letters")
+
                 inv_strs = self.last_observation[inv_strs_index]
-                for line in inv_strs:
+                inv_letters = self.last_observation[inv_letters_index]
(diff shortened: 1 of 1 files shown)


nle/env/base.py
```diff
@@ -477,12 +477,17 @@ def render(self, mode="human"):
...
                 inv_strs = self.last_observation[inv_strs_index]
-                for line in inv_strs:
+                inv_letters = self.last_observation[inv_letters_index]
+                for letter, line in zip(inv_letters, inv_strs):
                     if np.all(line == 0):
                         break
-                    print(line.tobytes().decode("utf-8"))
-            except ValueError:  # inv_strs not used.
+                    print(
+                        letter.tobytes().decode("utf-8"), line.tobytes().decode("utf-8")
```
heiner, line 488: I'm almost disappointed this is not `print(*(a.tobytes().decode("utf-8") for a in (letter, line)))`.

Almost, but not really.

heiner (approved):

nle/env/base.py
yobibyte: How about ```print(*map(lambda x: x.tobytes().decode('utf-8'), (letter, line)))```?

heiner: It just gets worse :D

PR #101, opened 2020-10-23
Add TELEPORT to non-RL actions.

yobibyte:


heiner:
I'm fine to put this in for now but note that there's other ways of getting the "select something on the screen" type of input (e.g., higher-level versions of fireball, cone of cold, and even more likely short-term, scrolls of stinking cloud).

It's probably a better idea to try to detect this situation in particular.

yobibyte:
Agree with the general approach, but not sure if this is the priority now and if we are anywhere close to use those.

heiner (approved):

heiner (approved):
LGTM.

Let's squash-commit this as I don't think we want or need the history in this case.

PR #108, opened 2020-11-03
Make NETHACKOPTIONS a tuple to protect from modification.

yobibyte:


nle/nethack/nethack.py
```diff
@@ -115,7 +115,7 @@ def __init__(
         shutil.copyfile(DLPATH, dlpath)
 
         if options is None:
-            options = NETHACKOPTIONS
+            options = list(NETHACKOPTIONS)
```
heiner, line 118: Is this `list` call necessary given the next line?

heiner (approved):
I like it. Let's wait for the tests to pass before submitting though.

nle/nethack/nethack.py
yobibyte: You're right!

heiner (approved):
LGTM.

PR #109, opened 2020-11-03
Remove temporary directory

heiner:
Thanks to @yobibyte for pointing out we don't do this; this used to happen but apparently I screwed it up.

Since libnethack cd's into vardir we have to get out of there again. Ideally, we should make libnethack not cd anywhere.

### DIFF
--- a/nle/nethack/nethack.py
+++ b/nle/nethack/nethack.py
@@ -100,7 +100,12 @@ class Nethack:
             )
 
         # Create a HACKDIR for us.
-        self._vardir = tempfile.mkdtemp(prefix="nle")
+        self._tempdir = tempfile.TemporaryDirectory(prefix="nle")
+        self._vardir = self._tempdir.name
+
+        # Save cwd and restore later. Currently libnethack changes
+        # directory on loading.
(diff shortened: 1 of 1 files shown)


nle/nethack/nethack.py
```diff
@@ -100,7 +100,12 @@ def __init__(
             )
 
         # Create a HACKDIR for us.
-        self._vardir = tempfile.mkdtemp(prefix="nle")
+        self._tempdir = tempfile.TemporaryDirectory(prefix="nle")
+        self._vardir = self._tempdir.name
```
yobibyte, line 104: tempdir._name might return AttributeError, not sure if we want to check that somehow. https://hg.python.org/cpython/file/3.6/Lib/tempfile.py#l725

yobibyte (approved):
Looks good, thank you!

nle/nethack/nethack.py
heiner: I think you might be confusing this with `TemporaryFile`: https://hg.python.org/cpython/file/3.6/Lib/tempfile.py#l790

PR #115, opened 2020-11-05
Fix a bad memory leak.

heiner:
It turns out I misunderstood what pybind11's py::array::ensure does.
Looking at the code superficially (https://github.com/pybind/pybind11/blob/master/include/pybind11/numpy.h#L785)
I assumed this function steals a reference, hence the call to
release(). But the reference is being stolen from the return value
of raw_array(), which returns a new reference. Hence ensure() itself
does not steal anything but in fact increases the refcount (and will
decref it in ~array).

Thanks to @yobibyte for both finding this issue and insisting that
it happens on the Python side (I thought this might be related to
malloc calls in NetHack itself).

### DIFF
--- a/win/rl/pynethack.cc
+++ b/win/rl/pynethack.cc
@@ -30,11 +30,11 @@ using namespace py::literals;
 
 template <typename T>
 T *
-checked_conversion(py::object obj, const std::vector<ssize_t> &shape)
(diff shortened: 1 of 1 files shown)


yobibyte (approved):
Thanks for fixing that, @heiner!

PR #116, opened 2020-11-05
Use same running Python version in CMake.

heiner:
In some rare cases, CMake can insist on using a different Python
version than the one that runs setup.py at the moment. This change
should prevent that.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,6 +21,7 @@ import sys
 import setuptools
 from setuptools.command import build_ext
 from distutils import spawn
+from distutils import sysconfig
 
 
 class CMakeBuild(build_ext.build_ext):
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -21,6 +21,7 @@
 import setuptools
 from setuptools.command import build_ext
 from distutils import spawn
+from distutils import sysconfig
```
condnsdmatters, line 24: Why distutils instead of Python 3.7s sysconfig?
```suggestion
import sysconfig
```

setup.py
```diff
@@ -53,6 +54,8 @@ def build_extension(self, ext):
             "-DCMAKE_INSTALL_PREFIX=%s" % sys.base_prefix,
             "-DCMAKE_LIBRARY_OUTPUT_DIRECTORY=%s" % output_path,
             "-DHACKDIR=%s" % hackdir_path,
+            "-DPYTHON_INCLUDE_DIR=%s" % sysconfig.get_python_inc(),
```
condnsdmatters, line 57:
```suggestion
            "-DPYTHON_INCLUDE_DIR=%s" % sysconfig.get_paths()['include'],
```

condnsdmatters (approved):
It seems fine, but I would consider using the Python version of sysconfig, which I assume does the same. The Python documentation seems to encourage people to discourage direct use of distutils...

setup.py
heiner: I don't want to require Python 3.7, `distutils` in setup.py seems less bad than in general Python and we are using it for `spawn` already.

PR #118, opened 2020-11-06
Fix extra newline at end of NLE run; fix fwrite when bz2 is disabled.

heiner:


src/nle.c
```diff
@@ -77,7 +77,7 @@ write_data(void *buf, int length)
     BZ2_bzWrite(&bzerror, nle->ttyrec_bz2, buf, length);
     assert(bzerror == BZ_OK);
 #else
-    assert(fwrite(buf, 1, length, nle->ttyrec) == 0);
+    assert(fwrite(buf, 1, length, nle->ttyrec) == length);
```
yobibyte, line 80: Can you explain this line to me?

yobibyte (approved):
Fix works for me (checked on fix-newline branch).

src/nle.c
heiner: This fixes a mistake I made earlier: `fwrite` doesn't signal an OK status with `0`, it returns the number of bytes written when everything went well.

This code path is only active when `NLE_BZ2_TTYRECS` is not defined, which I did to test out this change as it allows to use `nle-ttyplay -p`.

PR #119, opened 2020-11-06
Add terminal output as an observation.

condnsdmatters:
~Outstanding issues:~

~1. The rendering looks wrong. If you uncomment the test you will see that M's appear. Perhaps this is down to `wchar_t` sizes, or is there an off by some error? It feels like the cursor is moving around sometimes, but that should really change the state of the vt to an unexpected state.~ 

~2. Need to add the cmake instructions to find tmt in the right place. At locally i've just dunked the files in src and include.~

Edit: all sorted


condnsdmatters:
> Outstanding issues:
> 
>     1. The rendering looks wrong. If you uncomment the test you will see that M's appear. Perhaps this is down to `wchar_t` sizes, or is there an off by some error? It feels like the cursor is moving around sometimes, but that should really change the state of the vt to an unexpected state.
> 
>     2. Need to add the cmake instructions to find tmt in the right place. At locally i've just dunked the files in src and include.

Ready for review. The M's seem to come from misinterpreted reverse-linefeed characters. Forking libtmt and adding a controller is a fix that seems to work, but should be investigated more.

The CMake addition is crude, but maybe one source and header dont warrant their own 'library'. Happy to discuss.

win/rl/pynethack.cc
```diff
@@ -274,6 +283,8 @@ PYBIND11_MODULE(_pynethack, m)
     /* NetHack constants. */
     mn.attr("ROWNO") = py::int_(ROWNO);
     mn.attr("COLNO") = py::int_(COLNO);
+    mn.attr("NLE_TERM_LI") = py::int_(NLE_TERM_LI);
```
heiner, line 286: These should be the same as ROWNO at all times?

condnsdmatters: No - you need a top line for text and 2 bottom lines for states etc.

heiner: Ah right I confused it with LI and CO. But i feel we could use these names and be aligned with nethack and be more obscure to boot?

include/nle.h
```diff
@@ -7,6 +7,8 @@
 
 #include <fcontext/fcontext.h>
 
+#include "tmt.h"
```
heiner, line 10: We can _probably_ get by with a `typedef struct TMT TMT;` here and include `tmt.h` in nle.c only. I'm not a C guru but that strikes me as slightly preferable?

include/nleobs.h
```diff
@@ -18,16 +21,18 @@ typedef struct nle_observation {
...
+    short *inv_glyphs;       /* Size NLE_INVENTORY_SIZE */
     unsigned char
-        *inv_strs; /* NLE_INVENTORY_SIZE * NLE_INVENTORY_STR_LENGTH */
-    unsigned char *inv_letters;  /* NLE_INVENTORY_SIZE */
-    unsigned char *inv_oclasses; /* NLE_INVENTORY_SIZE */
+        *inv_strs; /* Size NLE_INVENTORY_SIZE * NLE_INVENTORY_STR_LENGTH */
+    unsigned char *inv_letters;  /* Size NLE_INVENTORY_SIZE */
+    unsigned char *inv_oclasses; /* Size NLE_INVENTORY_SIZE */
     unsigned char *screen_descriptions;  /* Size ROWNO * (COLNO - 1) * NLE_SCREEN_DESCRIPTION_LENGTH */
+    unsigned char *terminal_chars;       /* Size NLE_TERM_LI * NLE_TERM_CO */
```
heiner, line 34: I know I keep arguing about observation names but what do you think about "tty_chars"?

nle/nethack/nethack.py
```diff
@@ -11,7 +11,7 @@
 
 DLPATH = os.path.join(os.path.dirname(_pynethack.__file__), "libnethack.so")
 
-# TODO: Consider getting this from C++.
+# TODO: Consider getting this from C++. @EH
```
heiner, line 14: Not sure this TODO I put in is super relevant btw. We could easily expose and use the relevant NetHack macros here, but:

* This is _very_ static data, and
* Reading the numbers as plain numbers has some upsides.

BTW my Google training makes me go for `TODO(ghusername)`, although ideally we'd even to `TODO(#issueNo)`.

nle/tests/test_envs.py
```diff
@@ -49,6 +49,10 @@ def rollout_env(env, max_rollout_len):
     return reward
 
 
+def term_screen(obs):
+    return "\n".join("".join(chr(c) for c in row) for row in obs["terminal_chars"])
```
heiner, line 53: It's so weird that numpy/Python won't let us convert a uint8 array into bytes/strings more easily than that.

src/nle.c
```diff
@@ -26,8 +26,68 @@
 
 extern int unixmain(int, char **);
 
+short
+vt_font_attr_convert(TMTCHAR *c)
```
heiner, line 30: While this is a nice idea, I'd opt to not go for this, as:

* NetHack itself already has an opinion for how to turn this thing you call "font" into a bitset.
* For ML purposes, bitsets are not too useful. Everything that can be enumerated should be -- and the colours for instance can.

Note that "bold" doesn't actually mean bold. Instead "bold" controls the "bright" bit of colors, e.g. doubles the value the foreground can have from 8 to 16.

So my suggestion would be to have a "tty_colors" observation that for "most frames" aligns with the "colors" observation (that contains only the dungeon, which may get occluded in the tty output).

For the "specials", NetHack uses these bitset values:

```
                        MG_CORPSE	0x01
			MG_INVIS	0x02
			MG_DETECT	0x04
			MG_PET		0x08
			MG_RIDDEN	0x10
			MG_STATUE	0x20
			MG_OBJPILE	0x40
```

We can't infer this from the terminal output here. However, NetHack also won't use most of what you include here: E.g., I don't think it will change the background color (except via `reverse`). It will not use blink or underline and likely won't use `invisible` either. As I said, `bold` is a "color" property. So is "dim" I think, which stands for "character is half-bright". So in effect I believe (may be slightly wrong) the only thing we get in addition is a single bit for "reverse", which we could include in the colors, e.g. as a sign bit in a int8?

src/nle.c
```diff
@@ -26,8 +26,68 @@
...
+        }
+        tmt_clean(vt);
+        break;
+
+    case TMT_MSG_ANSWER:
+        break;
+
+    case TMT_MSG_MOVED:
+        break;
+    }
```
heiner, line 86: I believe (and compiler warnings should agree) that you're missing `TMT_MSG_CURSOR` here. I think its only function is to toggle cursor visibility which NetHack probably doesn't do, but it's cleaner to include it.

src/nle.c
```diff
@@ -26,8 +26,68 @@
...
+                }
+            }
+        }
+        tmt_clean(vt);
+        break;
+
+    case TMT_MSG_ANSWER:
+        break;
+
+    case TMT_MSG_MOVED:
```
heiner, line 84: This reminds me: Cursor position is something we'll likely use in the ttyrecs. It might be a good idea to include tty cursor position in the observations here, too. Either as 2-dim short `tty_cursor` array, or perhaps we create a new "other interesting stuff" observation -- something that the kitchen-sink `internal` does right now, which now also contains `uhunger` etc. We could refactor that to be a bit less random?

heiner (commented):
I like it. Small comments below. Biggest change I'd argue for:

* Turn "font" observation into basically-color.
* Rename observations to `tty_*`.

src/nle.c
rockt: I agree, adding the cursor position should be quite useful.

nle/nethack/nethack.py
condnsdmatters: Sorry I actually just put this in as a reminder to do it for this PR! I wouldnt normally put it in like this - it avoids some of the useful highlighting you can get from some editors

src/nle.c
condnsdmatters: added a cursor observation and update it on the TMT_MSG_MOVED case, which corresponds to cursor move. I haven't included a toggle display bool though, as per heiner's comment.

src/nle.c
condnsdmatters: added

src/nle.c
condnsdmatters: cool, i have changed font to color :) and we now mimic the info in `colors.h`

nle/nethack/nethack.py
condnsdmatters: so i went ahead and changed this to actually get the values from C++. let me know if you think i should change back.

condnsdmatters:
Some notes on profiling, given PR #130  to explain why I added commit 8aeab0d.

Below is old master:
old_master

And the changes before commit 8aeabd (ie terminal always emulated, costing about 2ksps):
oldtty

This commit restores the performance: 
tty

nle/tests/test_nethack.py
```diff
@@ -384,3 +384,41 @@ def test_glance_descriptions(self):
...
+        terminal_shape = tty_chars.shape
+        assert _pynethack.nethack.NLE_TERM_LI == terminal_shape[0]
+        assert _pynethack.nethack.NLE_TERM_CO == terminal_shape[1]
+
+    def test_observations(self):
+        game = nethack.Nethack()
+        game.reset()
+
+        tty_chars = game._obs_buffers["tty_chars"]
+        tty_colors = game._obs_buffers["tty_colors"]
```
heiner, line 408: We could actually compare a crop of this with the "colors" observation?

src/nle.c
```diff
@@ -26,8 +28,107 @@
...
+     * found standard in IBM color graphics, and are the same order as those
+     * found in src/color.h. We take the values from color.h, and choose
+     * default to be bright black (NO_COLOR) as nethack does.
+     *
+     * Finally we indicate whether the color is reverse, by indicating the
+     * sign
+     * of the final integer.
+     */
+    signed char color = 0;
+    switch (c->a.fg) {
```
heiner, line 44: There's no need to do it this way.

You can just do `(c->a.bold) ? BRIGHT | c->a.gf : c->a.gf` in all cases, where `BRIGHT` should be defined in NetHack and otherwise `#define BRIGHT 8`.

heiner (approved):
Looks good. Can we double-check the color logic? I'll ping you.

src/nle.c
condnsdmatters: TMT_COLOR_X does not line up with nethacks CLR_X - they are off by one.  As long as it is, this way is more explicit and probably more 'done properly'.

nle/tests/test_nethack.py
```diff
@@ -405,25 +412,37 @@ def test_new_observation_shapes(self):
...
             if chr(c) == "@":
                 assert font == 15  # BRIGHT_WHITE
             if chr(c) == " ":
-                assert font == 8  # NO_COLOR
+                assert font == 0  # NO_COLOR
+
+    def test_crop(self, game):
+        game.reset()
+
+        g_chars = game._obs_buffers["chars"].reshape(-1)
```
heiner, line 439: Perhaps a stupid question but did you print these to check if there even is any color here? :)

condnsdmatters: Do you mean - did I check that the values of g_cols and tty_cols aren't all black (0) and white (7)? Yep! Saw some 3's etc...

heiner: Awesome. Yes that's what I meant.

heiner (approved):
Lovely.

Issue #120, opened 2020-11-09
`tgetent` segfaults when using libncurses6


heiner:
## 🐛 Bug

```
$ gdb python
(gdb) r /opt/nle/nle/scripts/play.py
Starting program: /opt/conda/bin/python /opt/nle/nle/scripts/play.py
warning: Error disabling address space randomization: Operation not permitted
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7f3a1a881700 (LWP 3748)]
[New Thread 0x7f3a18080700 (LWP 3749)]
[New Thread 0x7f3a1787f700 (LWP 3750)]
Available actions: (<MiscAction.MORE: 13>, <CompassDirection.N: 107>, <CompassDirection.E: 108>, <CompassDirection.S: 106>, <CompassDirection.W: 104>, <CompassDirection.NE: 117>, <CompassDirection.SE: 110>, <CompassDirection.SW: 98>, <CompassDirection.NW: 121>, <CompassDirectionLonger.N: 75>, <CompassDirectionLonger.E: 76>, <CompassDirectionLonger.S: 74>, <CompassDirectionLonger.W: 72>, <CompassDirectionLonger.NE: 85>, <CompassDirectionLonger.SE: 78>, <CompassDirectionLonger.SW: 66>, <CompassDirectionLonger.NW: 89>, <MiscDirection.UP: 60>, <MiscDirection.DOWN: 62>, <MiscDirection.WAIT: 46>, <Command.KICK: 4>, <Command.EAT: 101>, <Command.SEARCH: 115>)

Thread 1 "python" received signal SIGSEGV, Segmentation fault.
0x00007f3a102959a9 in ?? () from /lib/x86_64-linux-gnu/libtinfo.so.6
(gdb) bt
#0  0x00007f3a102959a9 in ?? () from /lib/x86_64-linux-gnu/libtinfo.so.6
#1  0x00007f3a10295e45 in _nc_read_entry2 () from /lib/x86_64-linux-gnu/libtinfo.so.6
#2  0x00007f3a1028e9f6 in ?? () from /lib/x86_64-linux-gnu/libtinfo.so.6
#3  0x00007f3a1028ee05 in _nc_setupterm () from /lib/x86_64-linux-gnu/libtinfo.so.6
[...]

PR #121, opened 2020-11-09
Fix ncurses-related memory leak; various updates to help debugging

heiner:
Fix a termcap/terminfo/termlib/ncurses related memory leak.

Even after recent fixes, we were still seeing memory leaks. This
commit fixes another one.

The terminfo/termcap library is a weird 1980s mess. Among other
things, it cannot clean up after itself. The tgetent(3) call allocs
some memory; it also sets some global state that tells it to re-use
that memory in future calls (most programs will call tgetent only
once; NLE and screen/tmux might be exceptions).

However, or dlopen/dlclose trick dynamically loaded AND UNLOADED
libncurses, which probably reset its global state, creating new
memory allocations for each tgetent call. Or perhaps calling tgetent
several times just leaks in every case.

This change moves the tgetent call up one level into nledl.c, and
only does it once (per environment). This appears to fix any memory
issues I could see with valgrind (although ncurses will always show
up there) and appears to not make ./rlmain r leak on Ubuntu, measured
with

    pmap -x $!.

On MacOS, (1) valgrind appears to show a number of entries, but they all
look like false positives or (2) are dlopen/dlclose-related or (3)
are tgetent-via-nle_start related. The practial memory behavior on
MacOS is harder to ascertain. Measuring with

    ps -axm -o pid,%mem,rss,comm | grep rlmain
shows a slooow increase with occasional drops. Certainly slower
than before :D

An alternative to this change would be to rip out ncurses altogether,
e.g. by #undef'ing TERMLIB and #define'ing ANSI_DEFAULT. Doing so
creates other memory leaks in the hilites array in termcap.c which
can probably be fixed too.

### DIFF
--- a/sys/unix/nledl.c
+++ b/sys/unix/nledl.c
@@ -4,6 +4,9 @@
 #include <stdlib.h>
 #include <string.h>
 
+#include <curses.h>
+#include <term.h>
+
 #include "nledl.h"
 
 void
@@ -57,6 +60,16 @@ nle_ctx_t *
 nle_start(const char *dlpath, nle_obs *obs, FILE *ttyrec,
           nle_seeds_init_t *seed_init)
 {
+    static bool tc_started = FALSE;
+    if (!tc_started) {
+        const char *term = getenv("TERM");
+        if (!term || tgetent(NULL, term) < 1) {
+            fprintf(stderr, "Couldn't get terminal\n");
+            exit(EXIT_FAILURE);
+        }
+        tc_started = TRUE;
+    }
+
     /* TODO: Consider getting ttyrec path from caller? */
     struct nledl_ctx *nledl = malloc(sizeof(struct nledl_ctx));
     nledl->ttyrec = ttyrec;
--- a/win/tty/termcap.c
+++ b/win/tty/termcap.c
@@ -154,7 +154,7 @@ int *wid, *hgt;
 #endif
         TE = VS = VE = nullstr;
 #ifdef TEXTCOLOR
-        /* NLE: TODO(heiner): Re-enable and free these. */
+        /* NLE: TODO(heiner): Fix using ANSI_DEFAULT, free these. */
         for (i = 0; i < CLR_MAX / 2; i++)
             if (i != CLR_BLACK) {
                 hilites[i | BRIGHT] = (char *) alloc(sizeof("\033[1;3%dm"));
--- a/win/tty/termcap.c
+++ b/win/tty/termcap.c
@@ -154,6 +154,7 @@ int *wid, *hgt;
 #endif
         TE = VS = VE = nullstr;
 #ifdef TEXTCOLOR
+        /* NLE: TODO(heiner): Re-enable and free these. */
         for (i = 0; i < CLR_MAX / 2; i++)
             if (i != CLR_BLACK) {
                 hilites[i | BRIGHT] = (char *) alloc(sizeof("\033[1;3%dm"));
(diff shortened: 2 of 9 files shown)


yobibyte (approved):
LGTM.

win/tty/termcap.c
```diff
@@ -737,52 +735,10 @@ static const short tmspc10[] = { /* from termcap */
 #endif
 #endif
 
-/* delay 50 ms */
+/* NLE: Don't delay ever. */
```
condnsdmatters, line 738: Is this change somehow related to the bug fix? Or is it a drive-by perf win?

sys/unix/nledl.c
```diff
@@ -57,6 +60,16 @@ nle_ctx_t *
 nle_start(const char *dlpath, nle_obs *obs, FILE *ttyrec,
           nle_seeds_init_t *seed_init)
 {
+    static bool tc_started = FALSE;
+    if (!tc_started) {
+        const char *term = getenv("TERM");
+        if (!term || tgetent(NULL, term) < 1) {
```
condnsdmatters, line 66: Why are we sending in NULL here, instead of a buffer we can free?

sys/unix/nledl.c
```diff
@@ -57,6 +60,16 @@ nle_ctx_t *
 nle_start(const char *dlpath, nle_obs *obs, FILE *ttyrec,
           nle_seeds_init_t *seed_init)
 {
+    static bool tc_started = FALSE;
+    if (!tc_started) {
```
yobibyte, line 64: Why do we need this check?

condnsdmatters (approved):
Mostly cool, two small questions.

win/tty/termcap.c
yobibyte, line 738: Need for speed?

yobibyte (approved):

win/tty/termcap.c
heiner: Related in the sense that if one doesn't link against ncurses/terminfo/termcap by `#undef`'ing `TERMLIB`, there's symbols in `tty_delay_output` that the linker complains about.

Also we never called that function in the first place.

sys/unix/nledl.c
heiner: Welcome to the depth of various versions of termcap. Stallman's [doc on GNU termcap](https://www.gnu.org/software/termutils/manual/termcap-1.3/html_mono/termcap.html) says:

> If you are using the GNU version of termcap, you can alternatively ask tgetent to allocate enough space. Pass a null pointer for buffer, and tgetent itself allocates the storage using malloc. There is no way to get the address that was allocated, and you shouldn't try to free the storage.
>
> With the Unix version of termcap, you must allocate space for the description yourself and pass the address of the space as the argument buffer. There is no way you can tell how much space is needed, so the convention is to allocate a buffer 2048 characters long and assume that is enough. (Formerly the convention was to allocate 1024 characters and assume that was enough. But one day, for one kind of terminal, that was not enough.)
>
> No matter how the space to store the description has been obtained, termcap records its address internally for use when you later interrogate the description with tgetnum, tgetstr or tgetflag. If the buffer was allocated by termcap, it will be freed by termcap too if you call tgetent again. If the buffer was provided by you, you must make sure that its contents remain unchanged for as long as you still plan to interrogate the description.

(This is part of the reason termcap will always show up in memleak-checking tools like valgrind.)

However, GNU termcap isn't being used by modern distributions I believe. Instead, the manpage for [tgetent(3)](https://linux.die.net/man/3/tgetstr) says:

> The tgetent routine loads the entry for name. It returns 1 on success, 0 if there is no such entry, and -1 if the terminfo database could not be found. The emulation ignores the buffer pointer bp.

Perhaps NetHack tried to be compatible with an older / non-ncurses based version? However, they too never looked at the pointer they allocate (and it's totally unclear why that couldn't have been on the stack).

Long story short: This argument doesn't matter.

sys/unix/nledl.c
heiner: Otherwise we call `tgetent` several times again which we don't want to do.

win/tty/termcap.c
heiner: Yes (although we didn't call this function); but also this includes symbols that aren't present if `TERMLIB` isn't defined, and doesn't properly protect against that with `#ifdef`s.

heiner:
(Appears to have fixed #120.)

PR #122, opened 2020-11-09
Add uhunger to internal; fix eat task.

heiner:


include/nleobs.h
```diff
@@ -5,7 +5,7 @@
 #define NLE_MESSAGE_SIZE 256
 #define NLE_BLSTATS_SIZE 25
 #define NLE_PROGRAM_STATE_SIZE 6
-#define NLE_INTERNAL_SIZE 7
+#define NLE_INTERNAL_SIZE 8
```
rockt, line 8: What's that again?

rockt (approved):
Just did a brief check, looks good!

alexholdenmiller (approved):

include/nleobs.h
heiner: The size of the array in question. https://github.com/facebookresearch/nle/pull/122/files#diff-b995192c5d916483ad5aeee2cee68f00af99290788371bcc34be5c126b24366dR250 shows what the contents are. It grew by one because we added `uhunger`.

The `internal` observation is a bit of kitchen sink, or indeed the [Dalai Lama in a pizza store](https://www.youtube.com/watch?v=GogjFO8GNEo&ab_channel=nsotd2).

PR #127, opened 2020-11-13
Added architecture document and links in README.md.

condnsdmatters:
Added the architecture document that was found on Quip. It maybe more helpful to view the [rendered version](https://github.com/facebookresearch/nle/blob/eric/architecture/ARCHITECTURE.md) for reviewing purposes.


heiner (commented):
Looks good. My only suggestion would be:

(1) To move this file into doc/nle, and
(2) To reference it from README.md.

I'd like to avoid adding too many files to the root directory.

heiner:
Ah (2) is basically happening here, albeit with an extra link in between. We can probably link your document directly?

heiner (approved):
LGTM

PR #129, opened 2020-11-14
Fixup the test PyPI install workflow.

condnsdmatters:
The current `test_package.yml` workflow test has been broken for 4 months.  I assume it aims to test the nle packages from PyPI but it seems a bit confused, since there is a clone step in it.  Furthermore I cant tell if the purpose of this workflow changed, since the description of the key installation step changed from 'Install from repo in test mode" to  "Install nle via pip" a few months ago to accurately reflect the executed command. Perhaps this was a copy job from a workflow that tests installation from the repo?

At any rate the workflow was failing due to the fact that the cloned local `nle` was being accessed in the final step and it doesnt have the pynethack built, since nle was installed via pip.  Removing the cloned repo has the desired effect.

[Old test](https://github.com/facebookresearch/nle/actions/runs/362831941)
[New test (still fails due on 3.5 due to f strings)](https://github.com/facebookresearch/nle/actions/runs/363227992)

Btw to trigger this new workflow I had to add a section to the yaml, which should be visible in the intermittent PR history.

### DIFF
--- a/.github/workflows/test_package.yml
+++ b/.github/workflows/test_package.yml
@@ -4,6 +4,9 @@ name: Test nle PyPI package
 on:
   schedule:
     - cron: "0 6,18 * * *"
+  push:
+    branches:
+      - eric/workflows
 
 jobs:
   test_install:
--- a/.github/workflows/test_package.yml
+++ b/.github/workflows/test_package.yml
@@ -26,4 +26,5 @@ jobs:
     - name: Install nle via pip
       run: "pip install nle"
     - name: Check nethack is installed
-      run: "python -c 'import nle; import gym; gym.make(\"NetHack-v0\")'"
+      run: |
+        python -c 'import nle, gym; e = gym.make("NetHack-v0"); e.reset(); e.step(0)' 
--- a/.github/workflows/test_package.yml
+++ b/.github/workflows/test_package.yml
@@ -26,8 +26,6 @@ jobs:
     - name: Install dependencies
       run: |
         brew install cmake
-    - name: Clone repo
-      uses: actions/checkout@v2
     - name: Install nle via pip
       run: "pip install nle"
     - name: Check nethack is installed
--- a/.github/workflows/test_package.yml
+++ b/.github/workflows/test_package.yml
@@ -4,9 +4,6 @@ name: Test nle PyPI package
 on:
   schedule:
     - cron: "0 6,18 * * *"
-  push:
-    branches:
-      - eric/workflows
 
 jobs:
   test_install:
--- a/nle/nethack/nethack.py
+++ b/nle/nethack/nethack.py
@@ -154,7 +154,7 @@ class Nethack:
         # TODO ideally we need to check the validity of the requested items
         with open(os.path.join(self._vardir, WIZKIT_FNAME), "w") as f:
             for item in wizkit_items:
-                f.write("{}\n".format(item))
+                f.write("%s\n" % item)
 
     def reset(self, new_ttyrec=None, wizkit_items=None):
         if wizkit_items is not None:
--- a/nle/nethack/nethack.py
+++ b/nle/nethack/nethack.py
@@ -154,7 +154,7 @@ class Nethack:
         # TODO ideally we need to check the validity of the requested items
         with open(os.path.join(self._vardir, WIZKIT_FNAME), "w") as f:
             for item in wizkit_items:
-                f.write(f"{item}\n")
+                f.write("{}\n".format(item))
 
     def reset(self, new_ttyrec=None, wizkit_items=None):
         if wizkit_items is not None:


edran:
> I assume it aims to the test the nle packages from PyPI but it is seems a bit confused, since there is a clone step in it.

I'm 90% sure I wrote it this way because we weren't including the test files / requirements into the package.

edran (changes requested):
I would actually put this back, and run the tests from the cloned repo?

.github/workflows/test_package.yml
```diff
@@ -23,8 +23,6 @@ jobs:
     - name: Install dependencies
       run: |
         brew install cmake
-    - name: Clone repo
```
edran, line 26: I would actually put this back, and run the tests from the cloned repo?

edran (changes requested):

.github/workflows/test_package.yml
condnsdmatters: The only problem with doing this is that the versions will no longer match up. If you are installing from pip, you need to run the tests from that version, otherwise normal development will break this run regularly. 

In general if we aren't going to package tests, I would argue it is better to focus on catching bugs before shipping, and then just make sure the package can be installed and core functionality run. Perhaps we could add the a little to the bash script run instead?

condnsdmatters:
Also one more point - our README says we require Python 3.7+, but we have these tests that seem to suggest we support 3.5 and 3.6, even though [some of these are officially at the end of their life](https://www.python.org/downloads/release/python-3510/). 

I propose we either update the README or the tests we run.  Generally I prefer to give as much support as possible, but in this case I would say it really comes down to what y'all had in mind when writing the library.  It seems that everything tests fine on 3.6 up (based on our gh actions), and I think the only thing we might miss from 3.7 is that dicts are now in the language spec as ordered dicts by default (in 3.6 this is just an implementation detail for CPython). That said, I doubt we would rely on this anywhere. If we were to actively support 3.5 we'd also have to be careful not to use any f-strings.

LMK your thoughts and I can amend either the README or job entries here in this PR.

heiner:
Busy weekend huh?

Personally I like the idea of supporting older versions of Python and have been avoiding f strings for that reason. I'm not sure we need to promise we'll do that forever (or on more of a "best effort" basis) but if it comes practically for free right now, why not?

condnsdmatters:
> Busy weekend huh?
> 
> Personally I like the idea of supporting older versions of Python and have been avoiding f strings for that reason. I'm not sure we need to promise we'll do that forever (or on more of a "best effort" basis) but if it comes practically for free right now, why not?

Yeah its kind of nice for all the people who might be running Ubuntu 16.04 but cant upgrade the built in 3.5 Python for whatever reason.

nle/nethack/nethack.py
```diff
@@ -154,7 +154,7 @@ def _write_wizkit_file(self, wizkit_items):
         # TODO ideally we need to check the validity of the requested items
         with open(os.path.join(self._vardir, WIZKIT_FNAME), "w") as f:
             for item in wizkit_items:
-                f.write(f"{item}\n")
+                f.write("{}\n".format(item))
```
heiner, line 157: Idiosyncratic weird take  from @heiner: Please don't ever use `str.format`. It was a mistake in every sense. (Obviously the f string here was unnecessary too, that slipped through my review orf @yobibyte's code.)

In this case here, let's use either `f.write("%s\n" % item)` or indeed `f.write(item)` followed by `f.write("\n")`.

heiner (changes requested):

nle/nethack/nethack.py
rockt: What's bad about f strings? Slow?

condnsdmatters: isn't string.format considered the more pythonic, and the C-style printf considered old hat?

condnsdmatters: I remember moving across from some reason about the time I moved from py2.7 to 3. I think its to do with the fact that [its considered quirkier](https://docs.python.org/3/library/stdtypes.html#printf-style-string-formatting), and str.format was supposed to be formatting "done right".  

TBH i have no strong preferences either way because im not really aware of the great advantages either way, but I liked `str.format` for the named arguments that you can do. Very keen to hear whats behind the ideosyncratic take and am ultimately flexible.

condnsdmatters: @rockt It's because f strings are py3.6+ and I think the only thing stopping us supporting 3.5, which is default on ubuntu 16.04

edran: @rockt f-strings actually are great (and generally faster than other formatting systems in python), but they require python 3.6+.

My only issue with str.format is that they are fairly slow, and C-style formatting is slightly faster on average, so ¯\_(ツ)_/¯ (though f-strings a couple of times faster IIRC still).

heiner: f-strings are fine but not supported on older versions of Python 3.

`str.format` was just a mistake, made worse by a superstitious belief that it's the true / only way of doing this in Python3. `%` is better in almost every way. It can do named arguments as well if you want it to: `%(ham)s, %(spam)i" % dict(spam=42, ham="eggs")`.

condnsdmatters:
```suggestion
                f.write("%s\n" % item)
```

heiner: Fine by me :)

(Although for files it's perhaps ever so slightly more efficient to just write one string after another.)

condnsdmatters:
Alrighty @edran and @heiner got 2 changes requested, but i think I've addressed both of them. If you have any more LMK, otherwise I'll wait to merge until I get both approvals. 👍

heiner (approved):
Awesome, thanks!

edran (approved):

PR #130, opened 2020-11-15
Add benchmarks sps profile test.

condnsdmatters:
## TL:DR;
This PR involves 2 fixes. 
* 1st a fixup to `check_nethack_speed` to ensure it runs. Sadly as its not hooked into any automated tests I think it just got a bit out of sync with the API.
* 2nd is an introduction of a profiling test using pytest-benchmark. I think this set up will allow us to effectively and (relatively) reliably measure SPS for our code changes. The output of such a pytest-benchmark is very also informative - note the OPS column gives us the mean kilo-steps-per-second.  

## Example Output

master

## Why use the benchmark instead of the former script**? 

Well it seemed to me that the former script didn't really measure the steps per second of a single environment, rather the maximum throughput one could get running all the environments in suprocesses and putting the number of steps in a shared queue (which could be still useful!).

To illustrate my point consider the following screenshot, in which you choose 1, 2, 5, and (default) 10 environments to use. Clearly there's an communication overhead that becomes the dominant factor at 50k steps per second, which is misleading given each of the 10 environments can reach about 10k steps a second.

Screenshot 2020-11-15 at 15 46 03

As a result I propose using the pytest-benchmark to run performance tests, and perhaps rename `check_nethack_speed` to something more descriptive? I'm open to suggestions.

### DIFF
--- a/nle/scripts/check_nethack_speed.py
+++ b/nle/scripts/check_nethack_speed.py
@@ -28,7 +28,7 @@ def target(i, should_stop, queue):
 
 
 def play(should_stop, queue):
-    game = nethack.NetHack(archivefile=None)
+    game = nethack.Nethack()
 
     done = True
     steps = 0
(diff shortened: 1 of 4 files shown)


rockt:
That's great and really useful. Thanks @condnsdmatters !

heiner:
Black still not happy? (I can really recommend auto-running black on saving in your editor!)

Otherwise LGTM. This script is old and was used before we had libnethack. In a sense process overhead was an even worse problem at the time but perhaps we cared less :)

edran:
>Black still not happy? (I can really recommend auto-running black on saving in your editor!)

@condnsdmatters In the nle repo's root, run `$ pre-commit install` :)

edran (approved):
+1 on using `pytest-benchmark`. The added test looks good.

PR #139, opened 2020-12-04
Make space_dict global to reuse outside nle.

yobibyte:
I made it a tuple to protect the default object from modification.

### DIFF
--- a/nle/env/base.py
+++ b/nle/env/base.py
@@ -37,6 +37,99 @@ BLSTATS_SCORE_INDEX = 9
 
 SKIP_EXCEPTIONS = (b"eat", b"attack", b"direction?", b"pray")
 
+NLE_SPACE_ITEMS = (
+    (
+        "glyphs",
+        gym.spaces.Box(
+            low=0, high=nethack.MAX_GLYPH, **nethack.OBSERVATION_DESC["glyphs"]
+        ),
+    ),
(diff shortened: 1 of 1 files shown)


nle/env/base.py
```diff
@@ -37,6 +37,99 @@
 
 SKIP_EXCEPTIONS = (b"eat", b"attack", b"direction?", b"pray")
 
+NLE_SPACE_ITEMS = (
```
heiner, line 40: Any reason for not making this a dict already? Immutability?

yobibyte: Exactly. Modifying the default object seems dangerous.

nle/env/base.py
```diff
@@ -225,71 +319,7 @@ def __init__(
...
-                low=-15,
-                high=15,
-                **nethack.OBSERVATION_DESC["tty_colors"],
-            ),
-            "tty_cursor": gym.spaces.Box(
-                low=0, high=255, **nethack.OBSERVATION_DESC["tty_cursor"]
-            ),
-        }
-
+        space_dict = dict(NLE_SPACE_ITEMS) if space_dict is None else space_dict
```
heiner, line 322: I think this is fine as the two liner

```suggestion
        if space_dict is None:
            space_dict = dict(NLE_SPACE_ITEMS)
```

yobibyte: Wow, github allows committing suggestions.

heiner (approved):

PR #140, opened 2020-12-07
[WIP] Libnethack shared

tscmoo:
I'm creating this draft PR as things appear to be fully functional, but there are at minimum some cleanups necessary, and some more functionality I'd like to add.

This introduces a "shared" mode to the libnethack approach. It is Linux-only, and implements manual loading of the shared object file to avoid the need to copy the library anywhere and to better isolate the environment.
It is enabled by default on Linux and will fall back to the existing (dlopen) behavior on other platforms.
I called it shared as the shared object is actually shared between instances. Yey.

Some advantages:
  - *never* creates or writes to any files on the filesystem. NetHack is tricked into using temporary files in memory when it wants to write to something. Thus it doesn't leak anything even in the event of a crash, and instances are guaranteed not to inferere with each other (not even the leaderboard is shared). chdir is also supressed.
  - Should perform better, though this remains to be tested.
    - Resets should be much faster as they now involve little more than a 130kb memcpy.
    - Copies of the library are still done in memory, but they are copy-on-write mmapped copies - this allows them to share large portions of read-only instruction data, which in turn will be shared in the instruction cache in the CPU.
  - Some functions such as fork, popen, exit will currently throw an error, but my intention is for them to just end the episode. This allows us to continue even if nethack panics or quits for some reason.
  - It will be possible to track memory resources to detect leaks or force cleanup.

### DIFF
--- a/nle/nethack/nethack.py
+++ b/nle/nethack/nethack.py
@@ -75,13 +75,13 @@ def _set_env_vars(options, hackdir, wizkit=None):
     if wizkit is not None:
         os.environ["WIZKIT"] = wizkit
 
+_nhinstances = 0
 
 # TODO: Not thread-safe for many reasons.
 # TODO: On Linux, we could use dlmopen to use different linker namespaces,
 # which should allow several instances of this. On MacOS, that seems
 # a tough call.
 class Nethack:
-    _instances = 0
 
     def __init__(
         self,
@@ -93,7 +93,10 @@ class Nethack:
         wizard=False,
         hackdir=HACKDIR,
     ):
+        global _nhinstances
         self._copy = copy
+
+        _nhinstances = _nhinstances + 1
 
         if not os.path.exists(hackdir) or not os.path.exists(
             os.path.join(hackdir, "sysconf")
@@ -102,25 +105,44 @@ class Nethack:
                 "Couldn't find NetHack installation at '%s'." % hackdir
             )
 
-        # Create a HACKDIR for us.
-        self._tempdir = tempfile.TemporaryDirectory(prefix="nle")
-        self._vardir = self._tempdir.name
+        self.shared = False;
 
-        # Save cwd and restore later. Currently libnethack changes
-        # directory on loading.
-        self._oldcwd = os.getcwd()
+        if _pynethack.supports_shared():
+          # "shared" mode does some hacky things to enable using a
+          # shared libnethack.so, prevents writing to any files, and does
+          # not chdir.
+          self.shared = True
+          dlpath = DLPATH
+          self._hackdir = hackdir
+        else:
 
-        # Symlink a few files.
-        for fn in ["nhdat", "sysconf"]:
-            os.symlink(os.path.join(hackdir, fn), os.path.join(self._vardir, fn))
-        # Touch a few files.
-        for fn in ["perm", "logfile", "xlogfile"]:
-            os.close(os.open(os.path.join(self._vardir, fn), os.O_CREAT))
-        os.mkdir(os.path.join(self._vardir, "save"))
+          # Create a HACKDIR for us.
+          if os.getenv("SLURM_JOBID"):
+            self._tempdir = None
+            self._vardir = "/scratch/slurm_tmpdir/%s/nle-%d-%d" % (os.getenv("SLURM_JOBID"), os.getpid(), _nhinstances)
+            os.mkdir(self._vardir)
+          else:
+            self._tempdir = tempfile.TemporaryDirectory(prefix="nle")
+            self._vardir = self._tempdir.name
+          print("_vardir is ", self._vardir)
 
-        # Hacky AF: Copy our so into this directory to load several copies ...
-        dlpath = os.path.join(self._vardir, "libnethack.so")
-        shutil.copyfile(DLPATH, dlpath)
+          self._hackdir = self._vardir
+
+          # Save cwd and restore later. Currently libnethack changes
+          # directory on loading.
+          self._oldcwd = os.getcwd()
+
+          # Symlink a few files.
+          for fn in ["nhdat", "sysconf"]:
+              os.symlink(os.path.join(hackdir, fn), os.path.join(self._vardir, fn))
+          # Touch a few files.
+          for fn in ["perm", "logfile", "xlogfile"]:
+              os.close(os.open(os.path.join(self._vardir, fn), os.O_CREAT))
+          os.mkdir(os.path.join(self._vardir, "save"))
+
+          # Hacky AF: Copy our so into this directory to load several copies ...
+          dlpath = os.path.join(self._vardir, "libnethack.so")
+          shutil.copyfile(DLPATH, dlpath)
 
         if options is None:
             options = NETHACKOPTIONS
@@ -129,10 +151,10 @@ class Nethack:
             self._options.append("playmode:debug")
         self._wizard = wizard
 
-        _set_env_vars(self._options, self._vardir)
+        _set_env_vars(self._options, self._hackdir)
         self._ttyrec = ttyrec
 
-        self._pynethack = _pynethack.Nethack(dlpath, ttyrec)
+        self._pynethack = _pynethack.Nethack(dlpath, ttyrec, self.shared)
 
         self._obs_buffers = {}
 
@@ -154,6 +176,8 @@ class Nethack:
         return self._step_return(), self._pynethack.done()
 
     def _write_wizkit_file(self, wizkit_items):
+        if self._vardir is None:
+            raise RuntimeError("FIXME: shared wizkit: can't write to HACKDIR as it is a shared directory")
         # TODO ideally we need to check the validity of the requested items
         with open(os.path.join(self._vardir, WIZKIT_FNAME), "w") as f:
             for item in wizkit_items:
@@ -164,9 +188,9 @@ class Nethack:
             if not self._wizard:
                 raise ValueError("Set wizard=True to use the wizkit option.")
             self._write_wizkit_file(wizkit_items)
-            _set_env_vars(self._options, self._vardir, wizkit=WIZKIT_FNAME)
+            _set_env_vars(self._options, self._hackdir, wizkit=WIZKIT_FNAME)
         else:
-            _set_env_vars(self._options, self._vardir)
+            _set_env_vars(self._options, self._hackdir)
         if new_ttyrec is None:
             self._pynethack.reset()
         else:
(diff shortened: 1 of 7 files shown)


nle/nethack/nethack.py
```diff
@@ -93,34 +93,56 @@ def __init__(
...
+          self.shared = True
+          dlpath = DLPATH
+          self._hackdir = hackdir
+        else:
 
-        # Hacky AF: Copy our so into this directory to load several copies ...
-        dlpath = os.path.join(self._vardir, "libnethack.so")
-        shutil.copyfile(DLPATH, dlpath)
+          # Create a HACKDIR for us.
+          if os.getenv("SLURM_JOBID"):
```
tscmoo, line 120: This slurm stuff probably doesn't belong here and I'll most likely remove it from this PR

sys/unix/nleshared.cc
```diff
@@ -0,0 +1,707 @@
...
+    fclose(f);
+
+    hdr = (Elf64_Ehdr*)data.data();
+    boundsCheck(hdr);
+
+    baseaddr = ~0;
+    endaddr = 0;
+    size = 0;
+
+    forPh([&](Elf64_Phdr* ph) {;
```
heiner, line 58: I have a very vague notion of what this is doing: It's like a reimplementation of parts of Linux dynamic library loading. It interprets the data in the so in ELF format, creates manual copies ("forks"), and gives access to symbols a la `dlsym`.

I can't claim I understand much beyond that and perhaps a few comments might be good?

nle/nethack/nethack.py
```diff
@@ -75,13 +75,13 @@ def _set_env_vars(options, hackdir, wizkit=None):
     if wizkit is not None:
         os.environ["WIZKIT"] = wizkit
 
+_nhinstances = 0
 
 # TODO: Not thread-safe for many reasons.
 # TODO: On Linux, we could use dlmopen to use different linker namespaces,
```
heiner, line 81: We can probably update some of these comments when we merge this.

heiner (commented):
I'm not sure I can review this more than superficially.

Would be interested if this releves any lock contention issue the `dlopen`/`dlclose` cycles produced?

heiner:
Ah I read the PR description now.

> Some functions such as fork, popen, exit will currently throw an error, but my intention is for them to just end the episode. This allows us to continue even if nethack panics or quits for some reason.

If you make them call `nethack_exit(int status)` that should do it.

As for the not-writing any files approach: I'm not sure to what extend users are currently using the built-in highscore (written to a file, persistent for one instance of NLE). Probably not much, and very probably not enough to warrant doing something different than you are doing here.

If only there were a POSIX way of doing this :P

nle/nethack/nethack.py
```diff
@@ -154,6 +176,8 @@ def step(self, action):
         return self._step_return(), self._pynethack.done()
 
     def _write_wizkit_file(self, wizkit_items):
+        if self._vardir is None:
+            raise RuntimeError("FIXME: shared wizkit: can't write to HACKDIR as it is a shared directory")
```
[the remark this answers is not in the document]
tscmoo: Could, yes.
However, I am already using temporary in-memory files to trick nethack (using memfd_create), so I could instead provide an interface to let python write to those files directly, creating a bit of a virtual filesystem in memory (but very bare-bones). This wizkit file could thus live in memory.

nle/nethack/nethack.py
```diff
@@ -93,34 +93,56 @@ def __init__(
...
 
-        # Symlink a few files.
-        for fn in ["nhdat", "sysconf"]:
-            os.symlink(os.path.join(hackdir, fn), os.path.join(self._vardir, fn))
-        # Touch a few files.
-        for fn in ["perm", "logfile", "xlogfile"]:
-            os.close(os.open(os.path.join(self._vardir, fn), os.O_CREAT))
-        os.mkdir(os.path.join(self._vardir, "save"))
+        if _pynethack.supports_shared():
+          # "shared" mode does some hacky things to enable using a
```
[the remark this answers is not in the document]
tscmoo: Yes, it is. I apologize. I come from a world of 2 indentation. Formatting will be cleaned up.

sys/unix/nleshared.cc
```diff
@@ -0,0 +1,707 @@
+
+#if defined(__linux__) && defined(__x86_64__)
+
+#include <elf.h>
```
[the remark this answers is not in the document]
tscmoo: I am a huge fan of clang-format, but I don't have it set to run automatically.
I usually reorganize the commits and format everything before making the final PR

sys/unix/nledl.c
```diff
@@ -6,70 +6,102 @@
...
-
-    if (!nledl->dlhandle) {
-        fprintf(stderr, "%s\n", dlerror());
-        exit(EXIT_FAILURE);
-    }
-
-    dlerror(); /* Clear any existing error */
-
+#if defined(__linux__) && defined(__x86_64__)
+#define HASSHARED
```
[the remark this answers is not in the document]
tscmoo: There is still a runtime option of using dlopen etc even if the shared mode is builtin. I'd like to leave this possibility open just in case it doesn't work on someones computer. I can imagine some version of gcc, linker etc to produce a shared object that my code will not link successfully.

sys/unix/nleshared.cc
```diff
@@ -0,0 +1,707 @@
...
+  size_t finiarraysz;
+  NHLoader(std::string libPath, const std::unordered_map<std::string, void*>& overrides) {
+    FILE* f = fopen(libPath.c_str(), "rb");
+    if (!f) {
+      throw std::runtime_error("Failed to open '" + libPath + "' for reading");
+    }
+    fseek(f, 0, SEEK_END);
+    data.resize(ftell(f));
+    fseek(f, 0, SEEK_SET);
+    fread(data.data(), data.size(), 1, f);
```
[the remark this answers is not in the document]
tscmoo: Not in particular. I suppose it could be.

sys/unix/nleshared.cc
tscmoo: I will try to outline the behavior in comments.

tscmoo:
> If you make them call `nethack_exit(int status)` that should do it.

Great, thanks.

The records (highscore) file could be made persistent between instances, but is there even anything it could be used for?

> If only there were a POSIX way of doing this :P

The other approach to replacing these functions would of course be to do it in nethack, through the system specific configs and defines.

sys/unix/nleshared.cc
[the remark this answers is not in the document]
tscmoo: It is now!

nle/nethack/nethack.py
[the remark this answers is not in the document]
heiner: Cool makes sense.

sys/unix/nledl.c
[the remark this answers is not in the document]
heiner: Ah yes that makes sense.

sys/unix/nleshared.cc
```diff
@@ -0,0 +1,798 @@
...
+  static Mmap shared(const Fd& fd, size_t offset, size_t size, bool write) {
+    return make(fd, offset, size, write, true);
+  }
+  static Mmap private_(const Fd& fd, size_t offset, size_t size, bool write) {
+    return make(fd, offset, size, write, false);
+  }
+  operator bool() const noexcept {
+    return data_;
+  }
+  operator std::byte*() const noexcept {
```
heiner, line 97: (Not so much a review comment more of a C++ question:) Just for my own curiosity: What does this do? I couldn't find it in https://en.cppreference.com/w/cpp/types/byte

Also would `constexpr` make sense for these functions here now that we're doing C++17?

sys/unix/nleshared.cc
```diff
@@ -0,0 +1,798 @@
...
+  Fd(int fd) noexcept : fd(fd) {}
+  Fd(const Fd&) = delete;
+  Fd(Fd&& n) noexcept {
+    fd = std::exchange(n.fd, -1);
+  }
+  Fd& operator=(const Fd&) = delete;
+  Fd& operator=(Fd&& n) noexcept {
+    std::swap(fd, n.fd);
+    return *this;
+  }
```
heiner, line 38: I'm almost missing something with `::dup2` here :D

heiner (commented):
One more question:

> It will be possible to track memory resources to detect leaks or force cleanup.

How would you do that? Overriding each of `malloc`, `alloc`, `posix_memalign`, etc?

heiner:
> The records (highscore) file could be made persistent between resets, but is there even anything it could be used for?

It's not very important; the only effect a persistent highscore has is that the replays contain this relative comparison of the episode that just ended with other episodes. I'm not sure to what extend researchers look at replays and the highscore data in there.

sys/unix/nleshared.cc
tscmoo: It's just a cast operator so it can be converted to a byte*. It's not actually used anywhere, though.

I don't think `constexpr` makes sense as there is no way to mmap anything at compile-time, so this type has no use in a constexpr context.

sys/unix/nleshared.cc
tscmoo: > One more question:
> 
> > It will be possible to track memory resources to detect leaks or force cleanup.
> 
> How would you do that? Overriding each of `malloc`, `alloc`, `posix_memalign`, etc?

Yes, exactly.

sys/unix/nleshared.cc
heiner: That makes sense, thanks!

sys/unix/nleshared.cc
```diff
@@ -0,0 +1,798 @@
...
+
+    fileMem = Mmap::shared(diskFd, 0, diskFd.fileSize(), false);
+
+    hdr = (Elf64_Ehdr*)fileMem.data();
+    boundsCheck(hdr);
+
+    size_t baseaddr = ~0;
+    size_t endaddr = 0;
+
+    forPh([&](Elf64_Phdr* ph) {
```
heiner, line 156: I'm assuming this is fast enough to not need the ability to `break` based on e.g. a return value of the lambda?

tscmoo: This code only runs once in the entire lifespan of the process, so... There's really nothing to be gained by making this stuff faster.

nle/nethack/nethack.py
```diff
@@ -102,25 +100,38 @@ def __init__(
                 "Couldn't find NetHack installation at '%s'." % hackdir
             )
 
-        # Create a HACKDIR for us.
-        self._tempdir = tempfile.TemporaryDirectory(prefix="nle")
-        self._vardir = self._tempdir.name
+        self.shared = False
+
+        if _pynethack.supports_shared():
```
samvelyan, line 105: Can we also have a flag for enabling/disabling this shared functionality? 

This is for the scenario where we want to have a batch of environments each having its own nhdat file based on the level description file that each env in the batch is periodically updating?

PR #142, opened 2020-12-13
Rerange tty_colors to [0, 32)

tscmoo:
tty_colors had an unusual range perhaps intended to be [-16, 16), but ended up being... something strange.
This changes it to be [0, 32).
Seeing as nothing really uses this yet, it should be safe to change.


heiner (approved):
Thanks!

PR #145, opened 2021-01-08
pre-commit for python3.8

samvelyan:
If you install `nle` using the installation guide for the developer mode, then you are running into the following pre-commit error for black. This is because the conda environment in the guide is python3.8, not python3.7.

```
RuntimeError: failed to find interpreter for Builtin discover of python_spec='python3.7'
```

Updating the `.pre-commit-config.yaml` to python3.8 solves the problem.


condnsdmatters (approved):
LGTM

PR #147, opened 2021-02-12
Tty rendering for NLE environment

rockt:


nle/env/base.py
```diff
@@ -511,6 +532,32 @@ def get_seeds(self):
...
+            rendering += col_index_str + "\n"
+        for i in range(H):
+            line = []
+            for j in range(W):
+                char = tty_chars[i, j]
+                color = tty_colors[i, j]
+                line.append(COLORS[color] + chr(char))
+            row_index_str = COLORS[1] + str(i % 10)
+            if print_guides:
+                rendering += row_index_str + "".join(line) + row_index_str + "\n"
```
condnsdmatters, line 553: in general this string additions can get very slow because in Python string are immutable, and + is a binary operator.
so `"some long" + "strings" + "can be" + "expensive" + "to make" `

is evaluated as:
```
("some long" + "strings")  # all copied to "some long strings"
("some long string"+ "can be")  # copied to "some long strings can be"
("some long strings can be"  + "expensive")  ... # etc O(N^2)
```

Thats why the "Pythonic" thing is to collect the strings in a list and use `.join` method as you have above, which means the allocation is only done once (O(N))

nle/env/base.py
```diff
@@ -521,39 +568,8 @@ def render(self, mode="human"):
...
-                    # cf. termcap.c.
-                    start_color = "\033[%d" % bool(colors[r][c] & BRIGHT)
-                    start_color += ";3%d" % (colors[r][c] & ~BRIGHT)
-                    start_color += "m"
-
-                    print(start_color + chr(chars[r][c]), end=nh_HE)
-                print("")
-            return
+            obs = self.last_observation
+            return self.tty_render(obs, print_guides=False)
```
condnsdmatters, line 572: can I suggest either 
```suggestion
            self.tty_render(obs, print_guides=False)
            return 
```

or 

```suggestion
            image = self.tty_render(obs, print_guides=False)
            print(image)
            return 
```

The current API suggests render doesnt return anything, and so by writing it as either of these examples, we can clearly see that the printing must happen in the `tty_render` function, and not elsewhere in the code (ie perhaps with the returned value). 

in fact I would be tempted to go for the second example, remove the print function from the tty_render method, and rename it as "tty_to_string" or "get_tty_string". That way it can be programatically reused, possibly for logging by downstream users.

nle/env/base.py
```diff
@@ -131,6 +131,27 @@
 )
 
 
+# TODO: can probably be imported from somewhere?
+COLORS = [
```
condnsdmatters, line 135: I think this should work with all ansi compatible terminals, but Im not entirely sure. It could probably be clearer which colour is which as well... But also this might not be a priority.

condnsdmatters (commented):
Good and clear, some general comments on string addition, and api. Perhaps we could make terminal render mode a sepate 'mode' instead of simply deleting the old code, for users? I will see if we can get the ansi mapping from nethack object etc...

nle/env/base.py
condnsdmatters: TBF you have done the line.join for for the rows, so N in this case is only like 24 rather than 80, but its still an order of magnitude.

rockt:
Thanks @condnsdmatters — PTAL

nle/nethack/__init__.py
```diff
@@ -1,5 +1,6 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 from nle.nethack.actions import *  # noqa: F403
+
```
condnsdmatters, line 3:
```suggestion
```

nle/env/base.py
```diff
@@ -511,6 +532,38 @@ def get_seeds(self):
...
         return self.env.get_current_seeds()
 
+    def tty_render(self, chars, colors, print_guides=False):
+        # tty_chars_index = self._observation_keys.index(keys[0])
+        # tty_colors_index = self._observation_keys.index(keys[1])
+        # tty_chars = obs[tty_chars_index]
+        # tty_colors = obs[tty_colors_index]
+
+        tty_chars = chars
+        tty_colors = colors
```
condnsdmatters, line 542:
```suggestion
    def get_tty_rendering(self, tty_chars, tty_colors, print_guides=False):
```

nle/env/base.py
```diff
@@ -521,6 +574,14 @@ def render(self, mode="human"):
         """
         chars_index = self._observation_keys.index("chars")
         if mode == "human":
+            obs = self.last_observation
+            tty_chars_index = self._observation_keys.index("tty_chars")
+            tty_colors_index = self._observation_keys.index("tty_colors")
+            tty_chars = obs[tty_chars_index]
+            tty_colors = obs[tty_colors_index]
+            self.tty_render(tty_chars, tty_colors)
```
condnsdmatters, line 582:
```suggestion
            rendering = self.get_tty_rendering(tty_chars, tty_colors)
            print(rendering)
```

nle/env/base.py
```diff
@@ -511,6 +532,38 @@ def get_seeds(self):
...
+            row_index_str = COLORS[1] + str(i % 10)
+            if print_guides:
+                rendering.append(row_index_str + "".join(line) + row_index_str)
+            else:
+                rendering.append("".join(line))
+        if print_guides:
+            rendering.append(col_index_str)
+        rendering.append(COLORS[7])
+        rendering = "\n".join(rendering)
+        print(rendering)
```
condnsdmatters, line 564:
```suggestion
```

condnsdmatters: I suggest moving the printing to a calling function so we can reuse this function for say, logging,

nle/env/base.py
condnsdmatters: have moved this here because i think we wll want to reuse the conversion function in logging

condnsdmatters (approved):
Some nit pick suggestions, but looks good.

heiner:
Hey, would be interested in the history of this change. I think we should probably merge the "rendering" modes here, but I would first like to understand why we added this mode here?

PR #156, opened 2021-05-25
Eric/competition

condnsdmatters:
This is the branch that contains the Challenge Environment and the NeurIPS baseline we will release for it. 

The key differences from `NetHackChallenge` to `NetHackScore` are:
* All menus allowed
* Full key board action space allowed (with underlying code for quit, save and options switched off)
* Rotating characters (`characeter = '@'`)
*  `PREV_MSG` and `HISTORY` actions are removed (they seemed to be causing crashes)

The model itself we submit as baseline is based heavily on the model in Neurips2020 paper, and posted on the branch here (`neurips2020release`).  Here we have taken the best model and cut out all the other elements of the model, to keep the baseline simple, and to try to keep the model to only one file. 

The results of this run (for rotating characters and single characters are visible [here](https://fairwandb.org/nethack/competition3/reports/Starter-Kit--VmlldzoxMzk5) ). 

My suspicion is the model has gotten a bit slower (6000 sps) so there is probably room for a speed up. We can do this after the competition has started.

### DIFF
--- a/nle/env/base.py
+++ b/nle/env/base.py
@@ -94,7 +94,7 @@ NLE_SPACE_ITEMS = (
     ),
     (
         "inv_strs",
-        gym.spaces.Box(low=0, high=127, **nethack.OBSERVATION_DESC["inv_strs"]),
+        gym.spaces.Box(low=0, high=255, **nethack.OBSERVATION_DESC["inv_strs"]),
     ),
     (
         "inv_letters",
@@ -122,7 +122,7 @@ NLE_SPACE_ITEMS = (
         "tty_colors",
         gym.spaces.Box(
             low=0,
-            high=32,
+            high=31,
             **nethack.OBSERVATION_DESC["tty_colors"],
         ),
     ),
--- a/nle/env/base.py
+++ b/nle/env/base.py
@@ -116,13 +116,13 @@ NLE_SPACE_ITEMS = (
     ),
     (
         "tty_chars",
-        gym.spaces.Box(low=0, high=127, **nethack.OBSERVATION_DESC["tty_chars"]),
+        gym.spaces.Box(low=0, high=255, **nethack.OBSERVATION_DESC["tty_chars"]),
     ),
     (
         "tty_colors",
         gym.spaces.Box(
-            low=-15,
-            high=15,
+            low=0,
+            high=32,
             **nethack.OBSERVATION_DESC["tty_colors"],
         ),
     ),
@@ -210,6 +210,7 @@ class NLE(gym.Env):
         options=None,
         wizard=False,
         allow_all_yn_questions=False,
+        allow_all_modes=False,
         space_dict=None,
     ):
         """Constructs a new NLE environment.
@@ -235,11 +236,15 @@ class NLE(gym.Env):
                 If set to True, no y/n questions in step() are declined.
                 If set to False, only elements of SKIP_EXCEPTIONS are not declined.
                 Defaults to False.
+            allow_all_modes (bool):
+                If set to True, do not decline menus, text input or auto 'MORE'.
+                If set to False, only skip click through 'MORE' on death.
         """
 
         self.character = character
         self._max_episode_steps = max_episode_steps
         self._allow_all_yn_questions = allow_all_yn_questions
+        self._allow_all_modes = allow_all_modes
 
         if actions is None:
             actions = FULL_ACTIONS
@@ -339,6 +344,9 @@ class NLE(gym.Env):
         for a_idx, a in enumerate(self._actions):
             print(a_idx, a)
 
+    def _check_abort(self, observation):
+        return self._steps >= self._max_episode_steps
+
     def step(self, action: int):
         """Steps the environment.
 
@@ -360,15 +368,17 @@ class NLE(gym.Env):
         last_observation = tuple(a.copy() for a in self.last_observation)
 
         observation, done = self.env.step(self._actions[action])
-        observation, done = self._perform_known_steps(
-            observation, done, exceptions=True
-        )
+        is_game_over = observation[self._program_state_index][0] == 1
+        if is_game_over or not self._allow_all_modes:
+            observation, done = self._perform_known_steps(
+                observation, done, exceptions=True
+            )
 
         self._steps += 1
 
         self.last_observation = observation
 
-        if self._steps >= self._max_episode_steps:
+        if self._check_abort(observation):
             end_status = self.StepStatus.ABORTED
         else:
             end_status = self._is_episode_end(observation)
--- a/nle/nethack/actions.py
+++ b/nle/nethack/actions.py
@@ -14,9 +14,22 @@ def C(c):
     return 0x1F & c
 
 
-# Missing here:
-#   Some characters for text input (e.g., +).
-# General menu handling isn't part of this either.
+class TextCharacters(enum.IntEnum):
+    PLUS = ord("+")
+    MINUS = ord("-")
+    SPACE = ord(" ")
+    APOS = ord("'")
+    QUOTE = ord('"')
+    NUM_0 = ord("0")
+    NUM_1 = ord("1")
+    NUM_2 = ord("2")
+    NUM_3 = ord("3")
+    NUM_4 = ord("4")
+    NUM_5 = ord("5")
+    NUM_6 = ord("6")
+    NUM_7 = ord("7")
+    NUM_8 = ord("8")
+    NUM_9 = ord("9")
 
 
 class CompassCardinalDirection(enum.IntEnum):
@@ -76,6 +89,12 @@ class MiscAction(enum.IntEnum):
     MORE = ord("\r")  # read the next message
 
 
+class UnsafeActions(enum.IntEnum):
+    # currently these result in an error or undesirable behaviour
+    HELP = ord("?")  # give a help message
+    PREVMSG = C("p")  # view recent game messages
+
+
 class Command(enum.IntEnum):
     EXTCMD = ord("#")  # perform an extended command
     EXTLIST = M("?")  # list all extended commands
@@ -100,7 +119,6 @@ class Command(enum.IntEnum):
     FIGHT = ord("F")  # Prefix: force fight even if you don't see a monster
     FORCE = M("f")  # force a lock
     GLANCE = ord(";")  # show what type of thing a map symbol corresponds to
-    HELP = ord("?")  # give a help message
     HISTORY = ord("V")  # show long version and game history
     INVENTORY = ord("i")  # show your inventory
     INVENTTYPE = ord("I")  # inventory specific item types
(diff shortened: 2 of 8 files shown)


src/cmd.c
```diff
@@ -3351,6 +3351,26 @@ int final;
     en_win = WIN_ERR;
 }
 
+int nle_dosave() {
+    pline("You get the feeling there's only one way to save yourself..."); 
+    return 1;
+}
+
+int nle_done2() {
+    pline("You can't quit now, you're having so much fun!");
```
tscmoo, line 3360: I love these messages

nle/agent/core/vtrace.py
```diff
@@ -125,10 +125,7 @@ def from_importance_weights(
         vs = torch.add(vs_minus_v_xs, values)
 
         # Advantage for policy gradient.
-        broadcasted_bootstrap_values = torch.ones_like(vs[0]) * bootstrap_value
-        vs_t_plus_1 = torch.cat(
-            [vs[1:], broadcasted_bootstrap_values.unsqueeze(0)], dim=0
-        )
+        vs_t_plus_1 = torch.cat([vs[1:], torch.unsqueeze(bootstrap_value, 0)], dim=0)
```
tscmoo, line 128: What is the rationale for changing this file? Are you sure the broadcasting isn't needed?
Just wondering as this file is not original, but copied from elsewhere.

nle/nethack/actions.py
```diff
@@ -76,6 +89,12 @@ class MiscAction(enum.IntEnum):
     MORE = ord("\r")  # read the next message
 
 
+class UnsafeActions(enum.IntEnum):
+    # currently these result in an error or undesirable behaviour
+    HELP = ord("?")  # give a help message
+    PREVMSG = C("p")  # view recent game messages
```
tscmoo, line 95: So, is 'p' not in the action list?
Can this not be handled like quit, save, etc. below?

nle/scripts/play.py
```diff
@@ -100,10 +101,14 @@ def play(env, mode, ngames, max_steps, seeds, savedir, no_render, render_mode, d
...
-                    print("Previous action: %s" % repr(env._actions[action]))
-                env.render(render_mode)
+                # print('--------')
+                # print(f"Previous reward: {str(reward):64s}")
+                act_str = repr(env._actions[action]) if action is not None else ''
+                # print(f"Previous action: {str(act_str):64s}" )
+                # print('--------')
+                # env.render(render_mode)
+                # print('--------')
+                # print('\033[31A') # Go up 31 lines.
```
tscmoo, line 111: Are these lines supposed to be commented out? (also the ones below)

nle/nethack/actions.py
condnsdmatters: No - control+P isnt. Luckily we dont need this action at all, so thats fine.

nle/scripts/play.py
condnsdmatters: will adjust these

nle/agent/core/vtrace.py
condnsdmatters: https://github.com/facebookresearch/nle/blob/0424d298f52f21e4e3f7dda09381051ce8359e30/nle/agent/core/vtrace.py#L128

Pulled straight out of our `neurips2020release` branch

nle/scripts/play.py
condnsdmatters: done

condnsdmatters:
Although our CI is still down (and should be put up again asap) can confirm the code is all blacked, tests all pass, and the only flake8 errors we have are line too long in baseline model... (aren't these fine?)

condnsdmatters:
Changes:
* Quitting has been reenabled, to allow for safe clean up of NLE
* The competition environment now aborts if you go 10000 steps without advancing in game timer
* You can load the polyhydra in test mode (it was broken before)

Needless to say I could squash restructure these commits, and maybe should for quick reversion.

nle/agent/core/vtrace.py
heiner: Let's leave this file as-is.

nle/env/base.py
```diff
@@ -149,6 +149,22 @@
     "\x1b[95m",
     "\x1b[96m",
     "\x1b[97m",
+    "\033[30m",
```
heiner, line 152: What is this? Why do we have this constant?

nle/env/base.py
```diff
@@ -379,15 +403,18 @@ def step(self, action: int):
         last_observation = tuple(a.copy() for a in self.last_observation)
 
         observation, done = self.env.step(self._actions[action])
-        observation, done = self._perform_known_steps(
-            observation, done, exceptions=True
-        )
+        if (
+            not self._allow_all_modes or observation[self._program_state_index][0] == 1
```
heiner, line 407: What does this do? Doesn't seem very readable to me.

nle/env/base.py
```diff
@@ -379,15 +403,18 @@ def step(self, action: int):
...
+            observation, done = self._perform_known_steps(
+                observation, done, exceptions=True
+            )
 
         self._steps += 1
 
         self.last_observation = observation
 
-        if self._steps >= self._max_episode_steps:
+        if self._check_abort(observation):
```
heiner, line 417: Are we going to override _check_abort in subclasses?

heiner: Ah yes we do. Alright then.

nle/scripts/play.py
```diff
@@ -100,10 +112,15 @@ def play(env, mode, ngames, max_steps, seeds, savedir, no_render, render_mode, d
...
-                if action is not None:
-                    print("Previous action: %s" % repr(env._actions[action]))
+                print("--------")
+                print(f"Previous reward: {str(reward):64s}")
+                act_str = repr(env._actions[action]) if action is not None else ""
+                print(f"Previous action: {str(act_str):64s}")
+                print("--------")
                 env.render(render_mode)
+                print("--------")
+                if not print_all_frames:
```
heiner, line 122: That's a bit of a weird name for what it does (going back to override previous frames)?

nle/scripts/ttyplay.py
```diff
@@ -10,6 +10,7 @@
 import struct
 import termios
 import time
+from nle.nethack.actions import _ACTIONS_DICT
```
heiner, line 13: I don't like this as it makes ttyplay.py nethack-only all of a sudden. Better to add another script and import the relevant functions.

nle/tests/test_envs.py
```diff
@@ -365,7 +365,7 @@ def test_final_reward(self, env):
 
         # Hack to quit.
         env.env.step(nethack.M("q"))
-        _, reward, done, _ = env.step(env._actions.index(ord("y")))
+        obs, reward, done, _ = env.step(env._actions.index(ord("y")))
```
heiner, line 368: Why this change?

heiner (commented):
This polybeast version here truly is a beast, in the bad sense of the word. I'm not convinced hydra is adding anything here either, but I understand this is just a copy&paste from the agent repo.

I can see how this might not be the easiest setup to start experimenting.

nle/env/base.py
condnsdmatters: This checks if death the game is over. will improve clarity

nle/scripts/play.py
condnsdmatters: if you are _not_ printing all frames, then you are going back and overwriting instead.  i wanted a `store_true` parameter that would stop the overwriting behaviour. howabout `print_frames_separately`?

nle/scripts/ttyplay.py
condnsdmatters: do you mean `nle` only?

condnsdmatters: do you mean now you cant just take the script and play any old ttyrec, away from nle?

nle/tests/test_envs.py
condnsdmatters: i was undoing a previous change to the tests a few commits back, and slipped.

condnsdmatters:
```suggestion
        _, reward, done, _ = env.step(env._actions.index(ord("y")))
```

condnsdmatters:
I have rebased and restructured into 3 commits. I'm increasingly tempted to cut out the agent and move it to a branch - am happy to do this if we think this is sensible.

condnsdmatters:
All tests passed so rebased to include latest changes.

heiner:
> I have rebased and restructured into 3 commits. I'm increasingly tempted to cut out the agent and move it to a branch - am happy to do this if we think this is sensible.

I think moving the agent to a branch is a good idea.

condnsdmatters:
The agent code now lives at #163. When this is merged, I will rebase that one off merge so the versions all make sense (currently the "challenge" environment doesnt exist in that branch.

condnsdmatters:
(Also a quick drive by fix for the observation space bug spotted on Friday)

nle/nethack/actions.py
```diff
@@ -76,6 +89,12 @@ class MiscAction(enum.IntEnum):
...
+
+
 class Command(enum.IntEnum):
```
heiner, line 98: Previously, this was "all commands" (or something), and the "RL relevant ones" we gathered in their own enum. What's the philosophy now?

nle/scripts/play.py
```diff
@@ -67,7 +68,18 @@ def get_action(env, action_mode, is_raw_env):
...
+    env,
+    mode,
+    ngames,
+    max_steps,
+    seeds,
+    savedir,
+    no_render,
+    render_mode,
+    print_frames_separately,
+    **kwargs,
```
heiner, line 81: What do we use the **kwargs for?

nle/scripts/ttyplay.py
```diff
@@ -126,6 +135,12 @@ def process(f):
             continue
 
         if channel == 1:  # Input channel.
+            os.write(
+                1, b"\033[s\033[26;0f\033[37;1mFrame %d:\033[0m " % frame
+            )  # Save Cursor & Jump to L26
```
heiner, line 140: could this interfere with other saved cursors from the data stream?

heiner (approved):
Generally looks good to me.

nle/scripts/ttyplay.py
condnsdmatters: A quick google suggests, yes you can only save one cursor at a time, so yes this would interfere. On test rendering i've not seen any problems though for our ttyrec2s...

nle/scripts/play.py
condnsdmatters: the `debug` flag is created but does nothing in this function because it has been used previously. however this fn is called with `play(**vars(flags))`, so it needs a home here. the home i thought would be clearer in `**kwargs`, but i almost put `**_`

nle/nethack/actions.py
condnsdmatters: This enum wasn't really all commands because it didnt contain movement etc?  I figured if we were splitting out lets at least clearly mark the very dangerous ones.

PR #158, opened 2021-06-03
Change cmake install command.

heiner:
It seems a recent version of cmake (happened to me for cmake 3.20.0)
no longer installs nethackdir when called with
  cmake --build . --target install
Starting from cmake 3.15, the command
  cmake --install .
works instead (https://stackoverflow.com/a/56457739/1136208).

Fixes issues along the lines of
$ pip install -e .
$ python -m nle.scripts.play
FileNotFoundError: Couldn't find NetHack installation at '/Users/hnr/miniconda3/envs/extdev/lib/python3.8/site-packages/nle/nethackdir


condnsdmatters (approved):
LGTM

PR #160, opened 2021-06-04
Heiner/addhow

heiner:


win/rl/pynethack.cc
```diff
@@ -392,19 +420,20 @@ PYBIND11_MODULE(_pynethack, m)
...
-                     throw std::out_of_range(
-                         "Index should be between 0 and NUMMONS ("
-                         + std::to_string(NUMMONS) + ") but got "
-                         + std::to_string(index));
-                 v_h.value_ptr() = &mons[index];
-                 v_h.inst->owned = false;
-                 v_h.set_holder_constructed(true);
-             },
-             py::detail::is_new_style_constructor())
+        .def(
```
heiner, line 423: I have no idea why clang-format decided to change this.

I wish it was more stable version by version.

include/nleobs.h
```diff
@@ -16,6 +16,7 @@ typedef struct nle_observation {
     int action;
     int done;
     char in_normal_game;     /* Bool indicating if other obs are set. */
+    int how_done;            /* If game is really_done, how it ended. */
```
heiner, line 19: This could be a `game_end_types` but I think it's better to keep this file standalone.

src/end.c
```diff
@@ -1473,6 +1475,8 @@ int how;
         /* don't bother counting to see whether it should be plural */
     }
 
+    nle_done(how);
```
heiner, line 1478: We could add more info here, e.g., final score.

win/rl/pynethack.cc
```diff
@@ -212,6 +212,12 @@ class Nethack
         return obs_.in_normal_game;
     }
 
+    game_end_types
+    how_done()
+    {
+        return static_cast<game_end_types>(obs_.how_done);
```
heiner, line 218: This will be 0 (DIED) by default, even when the game hasn't finished yet. The way this is done here is that it has to be a member of `game_end_types`, which has no value for "not yet ended".

nle/nethack/nethack.py
```diff
@@ -224,3 +224,6 @@ def get_current_seeds(self):
 
     def in_normal_game(self):
         return self._pynethack.in_normal_game()
+
+    def how_done(self):
```
condnsdmatters, line 228: will this crash if its done outside of the end game?

src/nle.c
```diff
@@ -73,6 +73,8 @@ vt_char_color_extract(TMTCHAR *c)
     case (TMT_COLOR_WHITE):
         color = (c->a.bold) ? CLR_WHITE : CLR_GRAY; // c = 15:7
         break;
+    case (TMT_COLOR_MAX):
```
condnsdmatters, line 76: I think this is kind of wild but iirc TMT_COLOR_MAX.... i think is actually passed in sometime (by tmt)?!?! Really wild. I may be wrong but my some distant memory is recalling a really bonkers thing with tmt around here... maybe i used to have an assert i got rid of? i cant remember

condnsdmatters: is this a helpful comment? probably not.

win/rl/winrl.cc
```diff
@@ -256,6 +256,8 @@ NetHackRL::fill_obs(nle_obs *obs)
         obs->internal[5] = nle_seeds[0]; /* core */
         obs->internal[6] = nle_seeds[1]; /* disp */
         obs->internal[7] = u.uhunger;
+        obs->internal[8] =
```
condnsdmatters, line 259: We add u.urexp here, but we;re not using it anywhere?

condnsdmatters (approved):
This looks good to me and an important addition!

nle/nethack/nethack.py
heiner: No, but it will return `game_end_types.DIED` which corresponds to `0`.

src/nle.c
heiner: I added this b/c the compiler doesn't like not checking all enum values. If we should do something special here, let me know.

win/rl/winrl.cc
heiner: _If_ we at some point decided to use NetHack's "end of game" score, we would get it from here. See the commit message here https://github.com/facebookresearch/nle/pull/160/commits/c5f077c8cddb748ac9b25c5163a48f2a3987f590 for details.

win/rl/pynethack.cc
tscmoo: Just a heads up that you should install clang-format in your conda env (`conda install clang-format`) to get a recent version, since I've noticed system package managers seem to have very old versions.

PR #162, opened 2021-06-05
Fix a bad buffer overflow in inv_strs.

heiner:


condnsdmatters (approved):
We should check for this elsewhere. My bet is this could be causing problems in `screen_descriptions` and maybe also `message`.

PR #166, opened 2021-06-08
Change _reward_fn to take action taken

### LINKED ISSUE #155: Allow reward function to consider action taken
## 🚀  Feature: Allow reward function to consider action taken

It would be great if `_reward_fn` could take in the action taken, as well as the previous and current observation. This would enable users to define custom reward functions that take actions, which is common in RL literature. The context in which `_reward_fn` is called (`env.step(action)`) has this action in scope.

I think it would be relatively straightforward to implement by changing a few lines in `nle/env/base.py` and `nle/env/tasks.py`.

RobertKirk:
In standard RL often the reward function takes as input the action
taken. This change allows that to happen by adjusting _reward_fn.

See issue https://github.com/facebookresearch/nle/issues/155 for more details

Closes #155

### DIFF
--- a/nle/env/tasks.py
+++ b/nle/env/tasks.py
@@ -82,9 +82,11 @@ class NetHackScore(base.NLE):
         penalty += (new_time - old_time) * self.penalty_time
         return penalty
 
-    def _reward_fn(self, last_observation, observation, end_status):
+    def _reward_fn(self, last_observation, action, observation, end_status):
         """Score delta, but with added a state loop penalty."""
-        score_diff = super()._reward_fn(last_observation, observation, end_status)
+        score_diff = super()._reward_fn(
+            last_observation, action, observation, end_status
+        )
         time_penalty = self._get_time_penalty(last_observation, observation)
         return score_diff + time_penalty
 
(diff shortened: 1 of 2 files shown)


nle/env/tasks.py
```diff
@@ -194,7 +196,7 @@ def __init__(self, *args, **kwargs):
 
         super().__init__(*args, options=options, **kwargs)
 
-    def _reward_fn(self, last_observation, observation, end_status):
+    def _reward_fn(self, last_observation, action, observation, end_status):
         """Difference between previous gold and new gold."""
         del end_status  # Unused
```
heiner, line 201: Thanks! To keep this in line with how this code was written so far, would you mind adding `del action` lines to make linters happy about `action` not being used in these functions at this point?

heiner (commented):
Looks good, short comment below.

nle/env/tasks.py
RobertKirk: Sure thing, I've made that change

heiner (approved):
Awesome, thanks!

PR #171, opened 2021-06-17
Flag for not generating a random monster if not specified in the des-file

samvelyan:
Added an environment variable $NH_NO_RAND_MON which if set disables [random monster generation after every game step](https://nethackwiki.com/wiki/Monster_creation#Random_generation). The flag also disables random monster/object placements in procedurally generated mazewalks.

The changes are within the NetHack source code.


samvelyan:
Proposing a different solution to the same feature in https://github.com/facebookresearch/nle/pull/191.

PR #172, opened 2021-06-18
Fix NetHackChallenge no_progress_timeout

tscmoo:
The internal counter needs to reset when reset() is called, otherwise as soon as it triggers it will also trigger immediately on turn 1 for the next games (breaking NLE entirely).

I've verified that the no progress timeout works as intended with this fix.


heiner (approved):
Oh boy. Thanks!

heiner:
@condnsdmatters FYI.

PR #173, opened 2021-06-19
Some fixes to NLE and the competition task

heiner:
This PR fixes our C++ lint test (which apparently never really worked and at any rate was broken for months now) and flat-out disables seeding in NLE. The seeding commit is rather defensive for purposes of the NLE competition. Afterwards, we can `#define NLE_ALLOW_SEEDING 1` to turn on the old behaviour.

### DIFF
--- a/include/config.h
+++ b/include/config.h
@@ -55,6 +55,7 @@
 #define RL_GRAPHICS  /* Reinforcement learning 'window' port */
 
 #define NO_SIGNAL
+#define NOSHELL
 
 /*
  * Define the default window system.  This should be one that is compiled
--- a/nle/env/tasks.py
+++ b/nle/env/tasks.py
@@ -341,6 +341,13 @@ class NetHackChallenge(NetHackScore):
         # If the in-game turn count doesn't change for 10_000 steps, we abort
         self.no_progress_timeout = no_progress_timeout
 
+        def f(*args, **kwargs):
+            raise RuntimeError("Should not try changing seeds")
+
+        self.env.set_initial_seeds = f
+        self.env.set_current_seeds = f
+        self.env.get_current_seeds = f
+
     def reset(self, *args, **kwargs):
         self._turns = None
         self._no_progress_count = 0
@@ -360,3 +367,6 @@ class NetHackChallenge(NetHackScore):
             self._steps >= self._max_episode_steps
             or self._no_progress_count >= self.no_progress_timeout
         )
+
+    def seed(self, core=None, disp=None, reseed=True):
+        raise RuntimeError("NetHackChallenge doesn't allow seed changes")
--- a/.github/workflows/lint_cc.yml
+++ b/.github/workflows/lint_cc.yml
@@ -20,4 +20,4 @@ jobs:
       - name: Clone NLE repo
         uses: actions/checkout@v2
       - name: Run clang
-        run: "./nle/scripts/run-clang-format -r win/rl"
+        run: "./nle/scripts/run-clang-format -r win/rl src/nle.c sys/unix/nledl.c include/nle.h include/nledl.h include/nleobs.h"
--- a/.github/workflows/lint_cc.yml
+++ b/.github/workflows/lint_cc.yml
@@ -5,10 +5,14 @@ on:
     branches:
       - master
     paths:
-      - "win/rl/**.{c,cc,cpp,h,hh,hpp}"
+      - "src/nle**.c"
+      - "include/nle**.h"
+      - "win/rl/**.cc"
   pull_request:
     paths:
-      - "win/rl/**.{c,cc,cpp,h,hh,hpp}"
+      - "src/nle**.c"
+      - "include/nle**.h"
+      - "win/rl/**.cc"
   schedule:
     - cron: "0 6,18 * * *"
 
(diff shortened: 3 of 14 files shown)


nle/nethack/actions.py
```diff
@@ -153,7 +153,15 @@ class Command(enum.IntEnum):
...
     SEEALL = ord("*")  # show all equipment in use
+    SEEAMULET = ord('"')  # show the amulet currently worn
+    SEEARMOR = ord("[")  # show the armor currently worn
+    SEEGOLD = ord("$")  # count your gold
+    SEERINGS = ord("=")  # show the ring(s) currently worn
+    SEESPELLS = ord("+")  # list and reorder known spells
+    SEETOOLS = ord("(")  # show the tools currently in use"
     SEETRAP = ord("^")  # show the type of adjacent trap
+    SEEWEAPON = ord(")")  # show the weapon currently wielded
+    SHELL = ord("!")  # do a shell escape (not enabled in NLE build)
```
condnsdmatters, line 164: Can we just leave this out 
```suggestion
```

nle/nethack/actions.py
```diff
@@ -15,11 +15,11 @@ def C(c):
 
 
 class TextCharacters(enum.IntEnum):
-    PLUS = ord("+")
+    PLUS = ord("+")  # Also SEESPELLS.
```
condnsdmatters, line 18: Seems silly to have the same actions twice. Anyone looking to "lookup" a key press will have trouble. Much cleaner to make sure the mapping is one-to-one. I would remove here.

nle/env/tasks.py
```diff
@@ -341,6 +341,13 @@ def __init__(
         # If the in-game turn count doesn't change for 10_000 steps, we abort
         self.no_progress_timeout = no_progress_timeout
 
+        def f(*args, **kwargs):
```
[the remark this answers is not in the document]
condnsdmatters: Ah i have seen below now that the compilation changes result in thrown exceptions in pynethack.cc so this really is window dressing, presumably to be _extra clear_ to competitors?

version.txt
```diff
@@ -1 +1 @@
-0.7.1
+0.7.2
```
condnsdmatters, line 1: Would do this in a different PR. we also need to possible change SAVEDIR etc

condnsdmatters (commented):
The code all looks fine except I wouldnt mark this for release yet. Some discussion points listed. The only think I'm thinking about is:  do we whole hog overhaul action space or make an incremental change? Another option (crazy) is to provide mapping to old action space in some way?

include/config.h
```diff
@@ -55,6 +55,7 @@
 #define RL_GRAPHICS  /* Reinforcement learning 'window' port */
 
 #define NO_SIGNAL
+#define NOSHELL
```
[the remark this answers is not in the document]
heiner: Makes it such that https://github.com/facebookresearch/nle/blob/ceee6396797c5fe00eac66aa909af48e5ee8b04d/include/unixconf.h#L316 doesn't define `SHELL` such that https://github.com/facebookresearch/nle/blob/ceee6396797c5fe00eac66aa909af48e5ee8b04d/sys/unix/unixunix.c#L267 doesn't check `sysopt.shellers` for whether the USER CAN EXECUTE A SHELL COMMAND but instead refuses it for all users.

nle/env/tasks.py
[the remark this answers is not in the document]
heiner: History of this change: I wanted to remove the Python functions that do seed setting, but unfortunately the pybind11 object's attributes are read-only so they cannot be changed this way.

I guess we could not do this now.

nle/nethack/actions.py
```diff
@@ -112,11 +112,11 @@ class Command(enum.IntEnum):
     DROP = ord("d")  # drop an item
     DROPTYPE = ord("D")  # drop specific item types
     EAT = ord("e")  # eat something
-    ESC = C("[")  # escape from the current query/action
     ENGRAVE = ord("E")  # engrave writing on the floor
     ENHANCE = M("e")  # advance or check weapon and spell skills
-    FIRE = ord("f")  # fire ammunition from quiver
+    ESC = C("[")  # escape from the current query/action
```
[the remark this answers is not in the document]
heiner: Agreed.

The change here is from sorting the entries, which they were initially and then I inserted `ESC` at slightly the wrong spot.

nle/nethack/actions.py
heiner: If we do that we also need to add `!` to the not-available ASCII characters in the action test this PR adds.

I guess this depends on whether we want to be able to name something with `!`. I guess we don't.

nle/nethack/actions.py
heiner: The lookup dict deals with that. I guess for symmetry reasons I found it nicer to have `PLUS` available given that `MINUS` is. I'm fine with either.

version.txt
heiner: Done.

heiner:
Updated PR to not touch actions for now.

nle/nethack/actions.py
heiner: (Removed action changes from PR.)

condnsdmatters (approved):

PR #174, opened 2021-06-21
Default savedir to None (and remove unused archivefile)

tscmoo:
savedir is a feature that will save ttyrecs of runs - this can be unexpected when training a model and may write a *huge* amount of unwanted files.

This changes the default to *not* save ttyrecs.

If you want ttyrecs to be saved, please specify savedir when instantiating the env from now on.

`archivefile` is a parameter that may have been used at some point, but is currently unused, so let's remove it.

### DIFF
--- a/nle/env/base.py
+++ b/nle/env/base.py
@@ -186,8 +186,7 @@ class NLE(gym.Env):
 
     def __init__(
         self,
-        savedir="",
-        archivefile="nethack.%(pid)i.%(time)s.zip",
+        savedir=None,
         character="mon-hum-neu-mal",
         max_episode_steps=5000,
         observation_keys=(
--- a/nle/env/base.py
+++ b/nle/env/base.py
@@ -187,6 +187,7 @@ class NLE(gym.Env):
     def __init__(
         self,
         savedir=None,
+        archivefile=None,
         character="mon-hum-neu-mal",
         max_episode_steps=5000,
         observation_keys=(
--- a/nle/agent/agent.py
+++ b/nle/agent/agent.py
@@ -430,7 +430,7 @@ def train(flags):  # pylint: disable=too-many-branches, too-many-statements
         logging.info("Not using CUDA.")
         flags.device = torch.device("cpu")
 
-    env = create_env(flags.env, archivefile=None)
+    env = create_env(flags.env)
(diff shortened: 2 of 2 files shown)


condnsdmatters (commented):
Generally sensible wrt savedir, but we need to communicate these changes to AIcrowd.

WRT archivefile: this will break the benchmark training we provide to contestants and if we go ahead, release should be timed to match update of model, action space, etc... 

In general, we should just consider that removing archivefile is a breaking change to NLE api - why are we so certain we don't want an archivefile that keeps track of stats from all the ttyrecs that env created? I think this would be useful  if you are generating 1000s of ttyrecs, so I think we should think twice before breaking the api, lest we need it again in future.

nle/env/base.py
```diff
@@ -186,8 +186,7 @@ class StepStatus(enum.IntEnum):
 
     def __init__(
         self,
-        savedir="",
-        archivefile="nethack.%(pid)i.%(time)s.zip",
```
heiner, line 190: After talking to Eric he convinced me to leave this here for now in order to not confuse participants too much.

Let's just remove savedir in this PR (and also the archivefile=None calls above, given that the argument gets ignored) and we'll deal with archivefile in a future PR.

tscmoo: Can archivefile be defaulted to None, too, then? Same argument as for savedir

heiner: Yes.

condnsdmatters:
@tscmoo  any chance of reverting the deletion of archive file? happy to accept if so

condnsdmatters:
@tscmoo  any chance of reverting the deletion of archive file? happy to accept if so

heiner (approved):
LGTM.

Let's clean this up when we can.

condnsdmatters:
Have done w @heiner, and will accept.

condnsdmatters (approved):

tscmoo:
Ah, sorry, forgot about this. I'll merge it once the tests pass.

condnsdmatters:
Flaky test passing is probably down to starting with 'female gnomish archaeol-More-'.. Rerunning test and will submit a test fix.

condnsdmatters:
Test fix is in #182

PR #176, opened 2021-06-23
Observations in 'glyphs', 'inv_glyphs' now fixed to correspond to description of an object class instead of the name of object class.

condnsdmatters:
Prior to this commit, the glyph id for a 'Scroll of Identify' always had the same id. 
After this commit, the glyph id for a scroll called 'KERNOD_WELL' always has the same id. 
This mirrors the behaviour of tiles (for builds that use them).

### DIFF
--- a/win/rl/winrl.cc
+++ b/win/rl/winrl.cc
@@ -63,6 +63,26 @@ std::deque<std::string> win_proc_calls;
 bool in_yn_function = false;
 bool in_getlin = false;
 
+// Glyphs provide instructions for windows to render the game (see display.h).
+// At the start of the game, descriptions and properties of the object classes
+// are shuffled (see o_init.c) while the glyphs pointing to these classes are
+// not. This means glyph observations would always identify a 'wand of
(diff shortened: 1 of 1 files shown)


win/rl/winrl.cc
```diff
@@ -454,8 +471,9 @@ NetHackRL::update_inventory_method()
     inventory_.clear();
 
     for (otmp = invent; otmp; otmp = otmp->nobj) {
+        int glyph = obj_to_glyph(otmp, rn2_on_display_rng);
         inventory_.emplace_back(
-            rl_inventory_item{ obj_to_glyph(otmp, rn2_on_display_rng),
+            rl_inventory_item{ shuffled_glyph(glyph),
```
heiner, line 476: It's fine to do `shuffled_glyph(obj_to_glyph(otmp, rn2_on_display_rng))`.

heiner (approved):
Looks good modulo the clang-formatting.

Perhaps could not use the extra `glyph` variable but fine either way.

Thanks for finding this!

PR #179, opened 2021-06-23
Use cmap_to_glyph(S_stone) as "unset" glyph, not 0 (which is also PM_GIANT_ANT).

heiner:


samvelyan (commented):
Looks good to me. I've updated the tileset functionality accordingly. Specifically, when mapping glyph_ids to tile_ids for rendering, I replace NO_GLYPH with "dark part of the room" glyph id (corresponds to black square tile).

heiner:
> Looks good to me. I've updated the tileset functionality accordingly. Specifically, when mapping glyph_ids to tile_ids for rendering, I replace NO_GLYPH with "dark part of the room" glyph id (corresponds to black square tile).

That's a good point. Looking at https://github.com/facebookresearch/nle/blob/master/src/display.c#L1587 that is what NetHack does itself, so we should rather go with that.

I updated this PR.

condnsdmatters (approved):
lgtm

PR #180, opened 2021-06-23
Create 'misc' observation with input mode indicators 

condnsdmatters:
Indicators are (yn function, getline or xwaitforspace). Add this observation to NetHackChallenge.

This observation is an array of ints that can later be expanded to more
data if necessary, although the current entries are booleans (0 or 1).

~Tests inbound.~ Tests shipped.


heiner (approved):
Modulo lint check, LGTM.

(Btw `pre-submit install` should give you such a check at pre-commit time.)

PR #186, opened 2021-06-25
Update blstats only at end of bot() round.

heiner:
Otherwise, new x, y locations and dlevel will show up on the turn
that says "You descend the stairs.--More--", introducing a difference
between tty_chars and blstats.

### DIFF
--- a/win/rl/winrl.cc
+++ b/win/rl/winrl.cc
@@ -215,6 +215,9 @@ class NetHackRL
     std::array<std::string, MAXBLSTATS> status_;
     long condition_bits_;
 
+    void update_blstats();
+    long blstats_[NLE_BLSTATS_SIZE];
+
     void player_selection_method();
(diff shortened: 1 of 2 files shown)


condnsdmatters (approved):
LGTM

win/rl/winrl.cc
```diff
@@ -554,9 +560,10 @@ NetHackRL::status_update_method(int fldidx, genericptr_t ptr, int,
         return;
 
     // Needs to be kept in sync with the switch statement in rl_status_update.
-    if (fldidx == BL_FLUSH || fldidx == BL_RESET)
+    if (fldidx == BL_FLUSH || fldidx == BL_RESET) {
```
heiner, line 563: For context on the logic here, see e.g. https://github.com/NetHack/NetHack/blob/b65c93cdff3f52483645b29fc68f4adeb39ccbf3/doc/window.doc#L422

PR #191, opened 2021-07-01
Flag for disabling random monster spawning

samvelyan:
Adding the `spawn_monsters` flag to NLE to be able to disable random monster spawning that aren't described in the level's des-file (see [this](https://nethackwiki.com/wiki/Monster_creation) for more info). The same flag is also added to the NetHack class (True by default in both places). Setting this flag off triggers the `set_spawn_monsters` function in `win/rl/pynethack.cc`, which updates the corresponding value in `nle_init_settings` (previously called `nle_seed_init`). This later updates an external variable declared in `nle.h` that is checked in `src/allmain.c` before random monster spawning logic.

All of this functionality is available if `NLE_ALLOW_CONTROL` macro is defined. I modified the name of `NLE_ALLOW_SEEDING` since now it controls more than just seeding.

The above-mentioned `NLE_ALLOW_CONTROL` macro is manually defined in this branch in case someone want to try it out. I'll comment it out in the final version of this PR.

### DIFF
--- a/include/nleobs.h
+++ b/include/nleobs.h
@@ -13,7 +13,7 @@
 #define NLE_TERM_CO 80
 #define NLE_TERM_LI 24
 
-/* #define NLE_ALLOW_SEEDING 1 */
+/* #define NLE_ALLOW_CONTROL 1 */
 
 typedef struct nle_observation {
     int action;
@@ -42,12 +42,12 @@ typedef struct nle_observation {
 } nle_obs;
 
 typedef struct {
-#ifdef NLE_ALLOW_SEEDING
+#ifdef NLE_ALLOW_CONTROL
     unsigned long seeds[2]; /* core, disp */
     char reseed; /* boolean: use NetHack's anti-TAS reseed mechanism? */
 #else
     int _dummy; /* empty struct has size 0 in C, size 1 in C++ */
 #endif
-} nle_seeds_init_t;
+} nle_init_settings_t;
 
 #endif /* NLEOBS_H */
--- a/include/nleobs.h
+++ b/include/nleobs.h
@@ -13,7 +13,7 @@
 #define NLE_TERM_CO 80
 #define NLE_TERM_LI 24
 
-/* #define NLE_ALLOW_CONTROL 1 */
+#define NLE_ALLOW_CONTROL 1
 
 typedef struct nle_observation {
     int action;
@@ -45,6 +45,8 @@ typedef struct {
 #ifdef NLE_ALLOW_CONTROL
     unsigned long seeds[2]; /* core, disp */
     char reseed; /* boolean: use NetHack's anti-TAS reseed mechanism? */
+    int spawn_monsters; /* boolean: spawn monsters randomly after every step
+                          with some probability */
 #else
     int _dummy; /* empty struct has size 0 in C, size 1 in C++ */
 #endif
--- a/include/nleobs.h
+++ b/include/nleobs.h
@@ -18,7 +18,7 @@
 typedef struct nle_observation {
     int action;
     int done;
-    char in_normal_game;     /* Bool indicatin:g if other obs are set. */
+    char in_normal_game;     /* Bool indicating if other obs are set. */
     int how_done;            /* If game is really_done, how it ended. */
     short *glyphs;           /* Size ROWNO * (COLNO - 1) */
     unsigned char *chars;    /* Size ROWNO * (COLNO - 1) */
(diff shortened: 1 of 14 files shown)


src/allmain.c
```diff
@@ -6,6 +6,7 @@
 /* various code that was replicated in *main.c */
 
 #include "hack.h"
+#include "nle.h"
```
heiner, line 9: Let's not include nle.h here.

Instead, let's move the `extern int nle_spawn_monsters;` line from nle.h into this file (with a comment).

src/allmain.c
```diff
@@ -119,13 +120,16 @@ boolean resuming;
...
 
                     /* occasionally add another monster; since this takes
-                       place after movement has been allotted, the new
-                       monster effectively loses its first turn */
-                    if (!rn2(u.uevent.udemigod ? 25
+                       place after movement has been allotted (unless turned
+                       off in nle), the new monster effectively loses its first
+                       turn */
+                    if (nle_spawn_monsters) {
+                        if (!rn2(u.uevent.udemigod ? 25
```
heiner, line 127: I would suggest merging these ifs and doing `if (nle_spawn_monsters && !rn2(u.uevent.udemigod ? 25` ...

win/rl/pynethack.cc
```diff
@@ -177,14 +177,25 @@ class Nethack
...
-#ifdef NLE_ALLOW_SEEDING
-        seed_init_.seeds[0] = core;
-        seed_init_.seeds[1] = disp;
-        seed_init_.reseed = reseed;
-        use_seed_init = true;
+#ifdef NLE_ALLOW_CONTROL
+        init_settings_.seeds[0] = core;
+        init_settings_.seeds[1] = disp;
+        init_settings_.reseed = reseed;
+        use_init_settings = true;
```
heiner, line 198: Ah, this is a bit dangerous I think. `use_seed_init` used to change the seed arguments form the default. We might have to change this a bit.

win/rl/pynethack.cc
```diff
@@ -177,14 +177,25 @@ class Nethack
         }
     }
 
+    void
+    set_spawn_monsters(bool spawn_monsters)
+    {
+#ifdef NLE_ALLOW_CONTROL
+        init_settings_.spawn_monsters = spawn_monsters;
+        use_init_settings = true;
```
heiner, line 185: I think this has the side effect of changing the way seeding works, which isn't necessarily what we want here.

nle/nethack/nethack.py
```diff
@@ -168,6 +170,8 @@ def reset(self, new_ttyrec=None, wizkit_items=None):
             _set_env_vars(self._options, self._vardir, wizkit=WIZKIT_FNAME)
         else:
             _set_env_vars(self._options, self._vardir)
+        if not self._spawn_monsters:
+            self._pynethack.set_spawn_monsters(False)
```
heiner, line 174: I think it's a better idea to make it such that the underlying `_pynethack` object is either started with spawning on or off, without the ability to change this during its lifetime.

samvelyan:
Just realised this messes up the seeding values. Specifically, when turning off monster spawning and not providing seed values, the corresponding fields in the `nle_init_settings` struct are initialised by defaults and the episode look exactly the same after every reset. This needs to be fixed.

heiner (changes requested):
While this generally looks good, I think my idea of merging this setting with the seeding setting creates more complexity than I anticipated. I think that is also the reason the tests fail: Not giving a nle_seeds_init vs giving an empty/default one is a big difference, in that the empty one initializes with a seed of 0.

To keep the previous behaviour we should probably go another route: Instead of adding the spawn setting to the existing struct, it seem we will have to create two structs and add an additional argument to `nle_start`.

I would also suggest fixing the spawn value at construction of the C++ PyNetHack object -- I don't see a big benefit from being able to change spawning on resets. Let me know if you see this differently? Not allowing it would make the logic on the Python side simpler, which I think is a decent benefit. (For seeding, we can't do that unfortunately.)

samvelyan (commented):
Applied all change requests.

include/nleobs.h
```diff
@@ -45,6 +45,8 @@ typedef struct {
 #ifdef NLE_ALLOW_SEEDING
     unsigned long seeds[2]; /* core, disp */
     char reseed; /* boolean: use NetHack's anti-TAS reseed mechanism? */
+    int spawn_monsters; /* boolean: spawn monsters randomly after every step
```
heiner, line 48: Did you mean for this to be in the #ifdef?

heiner (commented):
Looks good, two small comments below.

include/nleobs.h
samvelyan: Good catch. Forgot to remove this from the struct.

samvelyan (commented):
Applied both requests.

heiner (approved):
Looks good, thanks.

PR #196, opened 2021-07-08
Fix blstats, again, again

### LINKED ISSUE #195: On the master version, there's a per-time-step negative reward at every step
## 🐛 Bug

There's a per-time-step negative reward of -0.01 on every step, regardless of whether the step should have increased the step counter or not. Digging into the code, this is because the calculation of the time penalty (`_get_time_penalty` in `NetHackScore`) always calculates `self._frozen_steps` as >1, and hence using `self.penalty_mode == "constant"` givesn us a fixed per-timestep penalty.

I've printed out the `blstats` for every timestep, and it seems that `blstats[20]` is no longer tracking timesteps (and in fact I believe none of the other `blstats` are either).

## To Reproduce

1. Install current master branch `pip install git+https://github.com/facebookresearch/nle.git`
2. Try playing `NetHackScore`
3. Observe `-0.01` reward on most time-steps

## Expected behavior

There should be `0` reward unless a time-step doesn't advance the time counter, which most of them should

## Environment
```
NLE version: 0.7.2+0f20c90
PyTorch version: 1.8.1
Is debug build: No
CUDA used to build PyTorch: None

OS: Mac OSX 11.4
GCC version: Could not collect
CMake version: version 3.19.2

Python version: 3.8
Is CUDA available: No
CUDA runtime version: No CUDA
GPU models and configuration: No CUDA
Nvidia driver version: No CUDA
cuDNN version: No CUDA

Versions of relevant libraries:
[pip3] numpy==1.19.5
[pip3] torch==1.8.1
[conda] blas                      2.108                       mkl    conda-forge
[conda] blas-devel                3.9.0                     8_mkl    conda-forge
[conda] libblas                   3.9.0                     8_mkl    conda-forge
[conda] libcblas                  3.9.0                     8_mkl    conda-forge
[conda] liblapack                 3.9.0                     8_mkl    conda-forge
[conda] liblapacke                3.9.0                     8_mkl    conda-forge
[conda] mkl                       2020.4             h08c4f10_301    conda-forge
[conda] mkl-devel                 2020.4             h694c41f_302    conda-forge
[conda] mkl-include               2020.4             h08c4f10_301    conda-forge
[conda] pytorch                   1.8.1                   py3.8_0    pytorch
```

## Additional context

Looking at the code, I imagine it's this commit https://github.com/facebookresearch/nle/commit/d62db5c4dc31fb29f8f04c9a3433ad7a4addfc7a which caused the issue, as that changed how the blstats are updated (although I don't understand what the bug is).

heiner:
Write moves / time every normal step.

This fixes a bug introduced by dropping the "time" option.

Hopefully fixes #195.

### DIFF
--- a/nle/tests/test_nethack.py
+++ b/nle/tests/test_nethack.py
@@ -343,44 +343,37 @@ class TestNethackFunctionsAndConstants:
 class TestNethackGlanceObservation:
     @pytest.fixture
     def game(self):  # Make sure we close even on test failure.
-        g = nethack.Nethack(playername="MonkBot-mon-hum-neu-mal")
+        g = nethack.Nethack(
+            playername="MonkBot-mon-hum-neu-mal",
+            observation_keys=("screen_descriptions", "glyphs", "chars"),
+        )
         try:
             yield g
         finally:
             g.close()
 
     def test_new_observation_shapes(self, game):
-        game = nethack.Nethack()
-        game.reset()
+        screen_descriptions, glyphs, *_ = game.reset()
 
-        screen_description_buff = game._obs_buffers["screen_descriptions"]
-        glyph_buff = game._obs_buffers["glyphs"]
-        glance_shape = screen_description_buff.shape
-        assert glyph_buff.shape == glance_shape[:2]
-        assert _pynethack.nethack.NLE_SCREEN_DESCRIPTION_LENGTH == glance_shape[-1]
-        assert len(glance_shape) == 3
+        assert len(screen_descriptions.shape) == 3
+        rows, cols, descr_len = screen_descriptions.shape
+        assert tuple(glyphs.shape) == (rows, cols)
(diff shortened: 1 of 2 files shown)


samvelyan:
Thanks @heiner for the quick fix. I can confirm that the PR resolves the issue on the MiniHack side, where we originally noticed the issue.

condnsdmatters (approved):

heiner:
> Thanks @heiner for the quick fix. I can confirm that the PR resolves the issue on the MiniHack side, where we originally noticed the issue.

Thanks for noticing. I wish we could unittest this better. Perhaps that's a reason to turn seed manipulation on again and try out specific trajectories (although there's no guarantee as the RNG is still OS dependent ...)

condnsdmatters (approved):

PR #197, opened 2021-07-08
Make NLE only run bzip2 when necessary

heiner:
Also fix misc test again and make installation a bit quieter.

### DIFF
--- a/src/nle.c
+++ b/src/nle.c
@@ -140,13 +140,14 @@ init_nle(FILE *ttyrec, nle_obs *obs)
 {
     nle_ctx_t *nle = malloc(sizeof(nle_ctx_t));
 
-    assert(ttyrec != NULL);
     nle->ttyrec = ttyrec;
 
 #ifdef NLE_BZ2_TTYRECS
-    int bzerror;
-    nle->ttyrec_bz2 = BZ2_bzWriteOpen(&bzerror, ttyrec, 9, 0, 0);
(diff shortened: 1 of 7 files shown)


src/nle.c
```diff
@@ -184,21 +185,23 @@ mainloop(fcontext_transfer_t ctx_transfer)
 }
 
 boolean
-write_data(void *buf, int length)
+write_ttyrec_data(void *buf, int length)
 {
     nle_ctx_t *nle = current_nle_ctx;
+    if (nle->ttyrec) {
```
condnsdmatters, line 191: do we need this check here given we have it below in nle_fflush? I understand trying to be defensive, but maybe we should pick one location/level of abstraction at which to put the if...

condnsdmatters (approved):
Only one comment, surrounding a potential extra if statement we might not need

src/nle.c
heiner: Wait this is a weird place to ask this question -- if `nle->ttyrec` is null the `BZ2_bzWrite` call below will surely fail as `nle->ttyrec_bz2` won't be a valid bz2 file?

heiner: Ah sorry, now I see what you mean: This function `write_ttyrec_data` never gets called with a NULL `ttyrec`. Fair point!

heiner: Removed the `if`, sorry for being slow in understanding.

Issue #188, opened 2021-06-29
Include "Pop-up" windows in observation


paulkent-um:
## 🚀 Feature

Include pop-up windows (such as "Things that are here:", or the spellcasting menu) in the observation space, either under "message", or under a new key.

## Motivation

I'm developing a bot for the 2021 Nethack challenge. The prospect of having to figure out where in the raw terminal output pop-ups are drawn up is not pleasing. Currently, if I want to get the information in these pop-up windows, that is the only way.

## Pitch

Copy over the information in the pop-ups under the observation key "message".

## Alternatives

If the pop-up windows turn out to take more than 255 characters to convey in some cases, or if it's possible for a message and a pop-up window to appear simultaneously, you could instead add a new key for this purpose, and allot it a larger buffer than the normal message observation.

condnsdmatters:
Hi there @paulkent-um  - I'm afraid copying over the item to message is a no go for a number of reasons:
* Firstly the message buffer would almost certainly be too short and we would have to cut the popups. Implementing our own buffering system is also a no go, since it would break a pretty fundamental design that whatever keypress you send goes to the game. It would be a little crazy, and involve hacks throughout the stack.
* Secondly, menus are more complicated that just messages. There's selections that take place ('what to drop') and associated letters/shortcuts, and more.  Perhaps for your use case parsing a giant string might suit you, but be certain that it would suddenly lead to edge cases. Consider a dropping menu: When choosing what to drop, NH gives you sub headings as below:
<img width="656" alt="Screenshot 2021-07-12 at 18 14 09" src="https://user-images.githubusercontent.com/6815729/125329061-f56c7200-e33c-11eb-9fcd-8a5307536d8d.png">

These are rendered as spearate rows, some are selectable, some arent. I have no doubt this is not the ideal solution for you either.

I have opened up a draft pr #207, with a proposed solution, with a new observation key, and after some discussions I'll revert with decisions on how to proceed.

paulkent-um:
I've left a comment on #207. One thing I did want to say here is:

> These are rendered as spearate rows, some are selectable, some arent. I have no doubt this is not the ideal solution for you either.

Perhaps not ideal, but if it was up to me, I'd settle for it. It'd take some unpacking to make useful, but I could absolutely work with that, which is a lot more than I can say about the present version of NLE.

PR #207, opened 2021-07-12
Add menu_strs observations.

condnsdmatters:
This is an observation that renders windows that are menus that pop up.

This is a DRAFT to discuss the issues that may be present in #188.  A proposed solution could look like like this draft.

NetHack has a window type called `NHW_MENU` that is used to render small items of text and multichoice menus (both of them).  Example contexts are when casting spells, choosing what to drop, what to pick up and put down when opening chests, and also just descrtiptions of what is in the floor.

There are two types of menu windows: ones that have `menu_items` - which include a letter and may be 'selected' (like casting spells, picking up items etc); and onethats have `strings` - which are just small items of text (like descriptions of what is in a place).

I propose adding a new observation, with the capacity of extracting all the useful data from the `menu_items`, and then perhaps placing a flag in `misc` to indicated first - are we in a menu - and second - what type of menu is this?

Potential problems: 
* How many menus can you have simultaneously? Is one the max?
* Are there any use cases that this misses? This doesnt cover large sections of text, for instance.
* How does this affect performance?

I'm sure there are other questions, but this is a start.

Note: There is onle line where I readjust the size of the vectors of windows. For some reason we don't 'delete' a window when it is deleted, we just manually tombstone with null, which affects `size()` function on the vector. I assume theres a very good reason for this, since it seems a little counter intuitive, (we only do the resize when creating a new_window?). I have done the resize to get the code to work, but i would suggest just removing and adding items as usual if we can get there...


condnsdmatters:
Wow really fantastic - already a segfault SMH

condnsdmatters:
I tested this manually, but will dig into it tomorrow

paulkent-um:
Hey, OP of #188 here.

> Are there any use cases that this misses? This doesnt cover large sections of text, for instance.

As a matter of fact, there is at least one use case I can immediately think of that this misses. 

Since the agent doesn't choose for itself what race, role, and alignment to play, it's kind of important that the agent be able to figure out what was chosen for it, since each of those has substantial gameplay implications. The most obvious way to figure this out is through the "hello Agent and welcome to Nethack" message at the beginning of the game, but that text doesn't appear on the full moon or new moon (since the text noting the phase of the moon replaces it), so that's not a reliable solution. Perhaps the agent could figure out role by reading off the "agent the <rank>" straight from ttychars, or by reviewing their starting equipment, but race and alignment don't always have those tells. The only surefire way I know of to find out your race and alignment at game start is with the Attributes action (action 25, keypress 24)... which displays as a large section of text but not a menu.

condnsdmatters:
Hi @paulkent-um. Thanks for chiming in! 😄 

Before moving on, let me just refocus of this point of this PR - to serialize pop-up windows rather than anything else.  While I appreciate that determining the race and role of your player might be useful to you, if you are looking for a more direct route in to getting this info you should open a different issue and we can discuss there.

As it happens the output of the `attributes` command (Ctrl + X) *is* technically pop-up menu (`NHW_MENU`) of the `string` type mentioned above, and actually would be covered by this PR (I have tested it). However, here we uncover our first big problem. 

The NetHack game creates the window object, but it is up to implementation of the windowing system to decide how to display it if it is too big. This is not a problem for spells and items on the floor, but for `attributes` it is always too many rows for the screen and gets paginated.
 
In the case of this pagination, the current draft here would _always_ only return the first half of the window text in the numpy array, even when the screen is displaying the second half of the page.  This is clearly a bug, but hacks around this with some local pagination on the `winrl.cc` end seem an invitation to even more bugs. Thoughts from @heiner  welcome at this point.

condnsdmatters:
> Wow really fantastic - already a segfault SMH

BTW this segfault came from the bug mentioned above: (number of rows in the window > rows in the screen) -> SEGV

condnsdmatters:
image
Here are the profiling results.

Note, this nethack environment (`NetHack-v0`) isnt really doing anything that involves windows, but the copying will still be be taking place

win/rl/winrl.cc
```diff
@@ -290,6 +290,13 @@ NetHackRL::fill_obs(nle_obs *obs)
         obs->misc[0] = in_yn_function;
         obs->misc[1] = in_getlin;
         obs->misc[2] = xwaitingforspace;
+        obs->misc[3] = false;
```
heiner, line 293: Be careful with C++ booleans like `true/false` vs C `bools`. They can in some situations not be the same thing and map to objects that evaluate to different logical values (!)

In this case it's ok because the lhs is an int, so the type will just get casted. Given that, it might make sense to write an int here in the first place.

win/rl/winrl.cc
```diff
@@ -353,6 +360,50 @@ NetHackRL::fill_obs(nle_obs *obs)
...
+        bool found_menu = false;
+        for (int i = 0; i < windows_.size(); ++i) {
+            if (i != WIN_INVEN && windows_[i].get()->type == NHW_MENU) {
+                // We have a new menu window to be rendered on screen!
+                // The menu can either have 'rl_menu_items' in init, or
+                // simple text. In this case we expect only one of the two
+                // to be available, and this one goes into menu_strs
+                found_menu = true;
+                rl_window *win = windows_[i].get();
+                assert(!win->menu_items.empty() ^ !win->strings.empty());
```
heiner, line 374: Note that `assert`s are stripped out of non-debug builds.

win/rl/winrl.cc
```diff
@@ -353,6 +360,50 @@ NetHackRL::fill_obs(nle_obs *obs)
             std::memset(obs->message, 0, NLE_MESSAGE_SIZE);
         }
     }
+
+    if (obs->menu_strs) {
+        bool found_menu = false;
+        for (int i = 0; i < windows_.size(); ++i) {
+            if (i != WIN_INVEN && windows_[i].get()->type == NHW_MENU) {
```
heiner, line 367: I'll have to very carefully check the logic here. I'm especially surprised there's no `break` statement here?

Will get back to this.

win/rl/winrl.cc
```diff
@@ -670,6 +721,7 @@ NetHackRL::destroy_nhwindow_method(winid wid)
 {
     DEBUG_API("rl_destroy_nhwindow(wid=" << wid << ")" << std::endl);
     windows_[wid].reset(nullptr);
+    windows_.resize(wid);
```
heiner, line 724: This seems strange?

(1) It renders the line before needless.
(2) It assumes no index `i >= wid` is currently used?

heiner (commented):
I'll need to dig into this more to say something more useful here.

win/rl/winrl.cc
condnsdmatters: Yes for some reason I thought the windows were stored in a stack-style data structure (largely verified in practice) but maybe this is all wrong. Either way should probably go read more code.

win/rl/winrl.cc
condnsdmatters: Yes - I'm not sure what to do if there are two windows - so I didn't put a break while testing to see if I could trigger it. If we go with a solution like this we should have a break.

paulkent-um:
Hey, can I follow up on this? Do you know when/if I can expect this to be merged into the main branch?

PR #214, opened 2021-07-16
Change `stable` to `latest` tag for our docker images.

condnsdmatters:
Adding this tag means we can do `docker pull fairnle/nle-xenial` and the
`latest` tag will be selected, instead of having to specify `docker pull
fairnle/nle-xenial:v0.7.2` etc.


heiner (approved):
If this is better, I'm happy with it.

Issue #215, opened 2021-07-16
Make statuses like 'Blind', 'Conf', 'Hallu', 'Slimed' available in BLStats


tanderson11:
Certain character statuses in Nethack are made visible to the player on the bottom line such as hallucination, confusion, blindness, illness, being stunned. While these characters are legibile in the raw observation, they are not made available to the player via `nle` in the `BLStats` class. 

I think these statuses should be made available natively through `nle`. Playing Nethack in a terminal occasionally causes some of these statuses to be forced offscreen if the window is narrow or the player is suffering from many conditions. Personally, I would like to see this behavior be avoided and have the statuses made available regardless of window size or number of conditions, but I'd also be happy with other implementations as well.

heiner:
This has happened in #219 and was released in version 0.7.3.

PR #219, opened 2021-07-19
Add condition mask to blstats observation.

heiner:
Or'd combination of https://github.com/facebookresearch/nle/blob/master/include/botl.h#L57.

This implements #215.


heiner:
As a bitset, this isn't exactly ML-friendly.

One way of using this:

```
cond = blstats[25];
is_slimed = bool(cond & nethack.BL_MASK_SLIME)
```

heiner:
Added constants for the blstats indices, PTAL.

I'd rather not add 13 more entries to blstats tbh.

condnsdmatters (approved):
(y)

PR #220, opened 2021-07-20
Disabling seeding in NLE using an environment varariable

samvelyan:
Uses an environment variable called `NLE_DISABLE_SEEDING` to disable seeding in NLE. If it is not set (which is usually expected), seeding is enabled.

Here are the commands I used to test the functionality
```bash
$ pip install -e .
$ python -c 'import gym; import nle; env = gym.make("NetHackScore-v0"); env.reset(); print(env.seed(1))'
(1, 5336406323662859033, False)

$ export NLE_DISABLE_SEEDING=1
$ pip install -e .
$ python -c 'import gym; import nle; env = gym.make("NetHackScore-v0"); env.reset(); print(env.seed(1))'
Traceback (most recent call last):
File "<string>", line 1, in <module>
File "/Users/samvelyan/workspace/NLE_fb/nle/nle/env/base.py", line 533, in seed
self.env.set_initial_seeds(core, disp, reseed)
File "/Users/samvelyan/workspace/NLE_fb/nle/nle/nethack/nethack.py", line 191, in set_initial_seeds
self._pynethack.set_initial_seeds(core, disp, reseed)
RuntimeError: Seeding not enabled

$ ./nle/scripts/nh-clean-install
$ export NLE_DISABLE_SEEDING=0
$ pip install -e .
$ python -c 'import gym; import nle; env = gym.make("NetHackScore-v0"); env.reset(); print(env.seed(1))'
(1, 8242861833433268597, False)
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,6 +46,9 @@ class CMakeBuild(build_ext.build_ext):
 
         generator = "Ninja" if spawn.find_executable("ninja") else "Unix Makefiles"
 
+        use_seeding = os.environ.get("USE_SEEDING", "ON")
+        use_seeding = {"1": "ON", "0": "OFF"}.get(use_seeding, use_seeding.upper())
+
         cmake_cmd = [
             "cmake",
             str(source_path),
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -58,6 +61,12 @@ def build_extension(self, ext):
             "-DPYTHON_LIBRARY=%s" % sysconfig.get_config_var("LIBDIR"),
         ]
 
+        use_seeding = os.environ.get("USE_SEEDING", True)
+        if use_seeding == "OFF":
+            use_seeding = False
+        if not int(use_seeding):
```
samvelyan, line 67: I think, logically, if disabling seeding is `export USE_SEEDING=OFF`, enabling it back should be `export USE_SEEDING=ON`, which causes an error here when converting to int.

heiner: Fair point, fixed.

samvelyan: great thanks

heiner (approved):
Not ideal, but we can live with this solution.

Issue #222, opened 2021-07-21
Tests fail when seeding mode is on


samvelyan:
## 🐛 Bug

When the NLE_ALLOW_SEEDING is defined, seeding tests started to fail.

## To Reproduce

1.  `git clone https://github.com/facebookresearch/nle --recursive`
2. `cd nle`
3. open `include/nleobs.h` with an editor, and make the following change
```diff
-/* #define NLE_ALLOW_SEEDING 1 */
+#define NLE_ALLOW_SEEDING 1
```
4. `pip install -e .`
5. `python ~/anaconda3/envs/nle/bin/pytest -svx nle/tests/` 

An error such as follows happens *sometimes*, *not always*, so repeat if not observed initially

```
>               np.testing.assert_equal(obs0, obs1)
E               AssertionError:
E               Arrays are not equal
E               key='tty_chars'
E
E               Mismatched elements: 8 / 1920 (0.417%)
E               Max absolute difference: 255
E               Max relative difference: 2.55
E                x: array([[ 67,  97, 110, ...,  47, 114, 101],
E                      [ 99, 111, 114, ...,  32,  32,  32],
E                      [ 32,  32,  32, ...,  32,  32,  32],...
E                y: array([[ 67,  97, 110, ...,  47, 114, 101],
E                      [ 99, 111, 114, ...,  32,  32,  32],
E                      [ 32,  32,  32, ...,  32,  32,  32],...

.... /nle/tests/test_envs.py:74: AssertionError
------
FAILED nle/tests/test_envs.py::TestGymEnvRollout::test_seed_rollout_seeded_int[500-NetHackOracle-v0] - AssertionError:
```

## Expected behavior

No such Assertion error should happen. Another assertion error must happen in a different test: `TestNetHackChallenge::test_no_seed_setting`.

## Environment
```
NLE version: 0.7.2+4473c7f
PyTorch version: 1.8.0
Is debug build: No
CUDA used to build PyTorch: None

OS: Mac OSX 11.4
GCC version: Could not collect
CMake version: version 3.19.3

Python version: 3.8
Is CUDA available: No
CUDA runtime version: No CUDA
GPU models and configuration: No CUDA
Nvidia driver version: No CUDA
cuDNN version: No CUDA

Versions of relevant libraries:
[pip3] numpy==1.19.5
[pip3] torch==1.8.0
[conda] torch                     1.8.0                    pypi_0    pypi
```
## Additional context

[...]

PR #223, opened 2021-07-21
Create empty "record" file in NLE's HACKDIR.

heiner:
I wonder how this even worked before? At any rate this seems obviously
necessary.

Fixes  https://github.com/facebookresearch/nle/issues/222.


samvelyan (approved):
LGTM

PR #224, opened 2021-07-21
Add GH Actions shield/badge.

heiner:
Documentation at https://docs.github.com/en/actions/managing-workflow-runs/adding-a-workflow-status-badge


condnsdmatters (approved):
🚀

PR #226, opened 2021-07-22
Add wizard mode option to play.py, refactor rendering logic.

heiner:


nle/nethack/nethack.py
```diff
@@ -77,10 +77,24 @@ def _set_env_vars(options, hackdir, wizkit=None):
         os.environ["WIZKIT"] = os.path.join(hackdir, wizkit)
 
 
+def tty_render(chars, colors):
```
condnsdmatters, line 80: could you add a (optional) cursor tothis? I believe cursor can be rendered with an underline effect?

condnsdmatters: or  reverse?

nle/scripts/play.py
```diff
@@ -116,15 +117,13 @@ def play():
                 print("-" * 8)
                 print(obs["blstats"])
                 if not FLAGS.print_frames_separately:
-                    print("\033[33A")  # Go up 33 lines.
+                    print("\033[33A")  # Go back up.
```
condnsdmatters, line 120: Useful to keep the comment (and maybe even str.fromat the 33) so anyone hacking know which bitto edit to adjust that. I've done so a lot in past. 
```suggestion
                    print("\033[%dA" % (33))  # Go back up 33 lines.
```

nle/scripts/play.py
```diff
@@ -116,15 +117,13 @@ def play():
...
             else:
                 print("Previous action:", action)
-                _, chars, _, _, blstats, message, *_ = obs
-                msg = bytes(message)
-                print(msg[: msg.index(b"\0")])
-                for line in chars:
-                    print(line.tobytes().decode("utf-8"))
-                print(blstats)
+                obs = dict(zip(nle.nethack.OBSERVATION_DESC.keys(), obs))
+                print(nle.nethack.tty_render(obs["tty_chars"], obs["tty_colors"]))
```
condnsdmatters, line 124: You've cut message intentionally yes?

condnsdmatters (approved):

nle/scripts/play.py
heiner: Yes because `tty_chars` has that data.

nle/scripts/play.py
heiner: Happy to splice the `33` in like you suggest, but this is the comment that is just bound to go out of sync, which is extra bad here since there's also the `\033` here ...

nle/nethack/nethack.py
heiner: Nice idea. Done.

nle/scripts/play.py
heiner: Moved to make it more expressive by introducing a suggestive function.

nle/nethack/nethack.py
heiner: image

heiner: It underlines the `f` a little bit because the cat and the hero are so close.

nle/scripts/play.py
```diff
@@ -39,6 +39,10 @@ def no_echo():
         termios.tcsetattr(0, termios.TCSAFLUSH, tt)
 
 
+def go_back(num_lines):
```
condnsdmatters, line 42: 👍

nle/tests/test_nethack.py
```diff
@@ -508,3 +508,16 @@ def test_misc_wait_for_space(self, game):
         game.step(ord(" "))
         assert np.all(misc == 0)
         np.testing.assert_array_equal(misc, internal[1:4])
+
+
+class TestAuxillaryFunctions:
+    def test_tty_render(self):
+        text = ["DE", "HV"]
```
condnsdmatters, line 515: 👍 👍 👍 👍

nle/env/base.py
```diff
@@ -569,16 +569,18 @@ def render(self, mode="human"):
...
+            chars = self.last_observation[self._observation_keys.index("chars")]
+            colors = self.last_observation[self._observation_keys.index("colors")]
             print(nethack.tty_render(chars, colors))
             return
-        elif mode == "ansi":
-            chars = self.last_observation[chars_index]
+
+        if mode in ("ansi", "string"):  # Misnomer: This is the least ANSI of them all.
+            chars = self.last_observation[self._observation_keys.index("chars")]
+            # TODO: Why return a string here but print in the other branches?
```
condnsdmatters, line 580: Returning a string is useful for things like WANDB, loggers, etc i think

condnsdmatters (approved):

nle/env/base.py
heiner: But it kinda breaks the API (and pylint doesn't like it).

But okay.

PR #227, opened 2021-07-22
Add isort to pre-commit and lint_python.yml workflow, apply isort.

heiner:


nle/tests/test_nethack.py
```diff
@@ -5,11 +5,10 @@
 import warnings
 
 import numpy as np
-
 import pytest
 
-from nle import nethack, _pynethack
-
+from nle import _pynethack
+from nle import nethack
```
condnsdmatters, line 11: this is interesting ... no multiple imports from same package?

pyproject.toml
```diff
@@ -10,3 +10,27 @@ exclude = '''
   | win/rl/nethack_rl
 )/
 '''
+
+[tool.isort]
+force_single_line = true
```
condnsdmatters, line 15: ah.... I'm not _certain_ this is the best idea, but I'm certainly happy to try it.  If it gets unworkable its easy to revert

pyproject.toml
```diff
@@ -10,3 +10,27 @@ exclude = '''
...
+profile = "black"
```
condnsdmatters, line 16: I was literally wondering if this would play well with black

condnsdmatters (approved):

pyproject.toml
heiner: I don't like that for `__init__.py` files which is why excluded those. But otherwise I like this better, and it's demanded by https://google.github.io/styleguide/pyguide.html#313-imports-formatting

PR #228, opened 2021-07-26
Use memfd_create + /proc/self/fd/{FD} for temporary DL on Linux.

heiner:
If `memfd_create` is unavailable, try using `O_TMPFILE` via `tempfile.TemporaryFile`.

MacOS has no `memfd_create` or `O_TMPFILE`. Unfortunately, using `/dev/fd/{FD}` as an argument to `dlopen` doesn't work either after unlinking from the file system (see https://gist.github.com/heiner/5179ccaef983b318c1fc451d7efbe95d#file-tmp_memfs_mac-py). So let's copy instead and hope vardir gets properly deleted at some point.

### DIFF
--- a/nle/nethack/nethack.py
+++ b/nle/nethack/nethack.py
@@ -1,6 +1,7 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 import os
 import shutil
+import sys
 import tempfile
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


heiner:
Thanks to @tscmoo for suggesting this!

tscmoo (approved):

PR #249, opened 2021-09-17
Test and fix for "NLE with wrong HACKDIR after spawn"

heiner:


samvelyan (approved):
I've made the same change (HACKDIR) on the MiniHack side and installed NLE using this branch. I then copied the test and replaced the Score task with one of the MiniHack tasks. It passes with no problems.

PR #253, opened 2021-09-20
Tests for NLE when using two environments sequentially or in threads

heiner:
And misc. changes.

Turns out #254 only shows up on my local machine (?!).


heiner:
The second test here shows that an approach like https://gist.github.com/heiner/67955538cce6375dbdd6fa97ffd11ce9 can work. In my tests, this could make SPS for tasks with very short episodes up to 50% faster.

condnsdmatters (approved):

PR #261, opened 2021-09-24
Remove space_dict argument to NLE.

heiner:
(Should not be merged before the end of the NetHack competition.)

This removes the as-far-as-I-know unused `space_dict` argument to the `NLE` class.

For a brief discussion, see https://github.com/facebookresearch/nle/commit/e8597298fe91f8335c86e5cd2b932751248d3f48#commitcomment-56904804


samvelyan (approved):
Looks good. Doesn't break anything on the minihack's side.

PR #271, opened 2021-10-07
Alternative fix for gym 0.21 changes.

### LINKED ISSUE #272: Gym 0.21 breaks a bunch of stuff
## 🐛 Bug

See https://github.com/facebookresearch/nle/pull/269 and https://github.com/facebookresearch/nle/pull/271 for details.

heiner:
This is an alternative to https://github.com/facebookresearch/nle/pull/269 to fix #272.

----

The most recent version of gym wraps environments with the `OrderEnforcing` wrapper (https://github.com/openai/gym/blob/master/docs/wrappers.md) if not disabled. This breaks our [tests](https://github.com/facebookresearch/nle/pull/268/checks?check_run_id=3821174783), `play.py` and a number of other places. For instance, `env.seed()` with more than 1 argument now also fails due to [`gym.Wrapper.seed`](https://github.com/openai/gym/blob/master/gym/core.py#L300). Possible fixes:
* Use `env.unwrapped` when arguments to `seed` are given
* Disable the `OrderEnforcing` wrapper. This will break NLE when using earlier versions of the gym library unless we do per-version magic (as in this PR). We cannot test for the presence of `OrderEnforcing` as it's not exported into `gym.wrappers` (_sigh_).
* Adhere to the gym environment better by doing `seed(self, (core, disp, reseed))` instead of `seed(self, core=None, disp=None, reseed=False)`.

We also use `env._actions` a bunch of times, in tests and in `play.py`. We should fix that, and other accesses to `_underscored` variables.

### DIFF
--- a/nle/env/__init__.py
+++ b/nle/env/__init__.py
@@ -1,23 +1,34 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
+import gym
 from gym.envs import registration
 
 from nle.env.base import NLE, DUNGEON_SHAPE
 
-registration.register(id="NetHack-v0", entry_point="nle.env.base:NLE")
+_version = "v0"
 
-registration.register(id="NetHackScore-v0", entry_point="nle.env.tasks:NetHackScore")
-registration.register(
-    id="NetHackStaircase-v0", entry_point="nle.env.tasks:NetHackStaircase"
-)
-registration.register(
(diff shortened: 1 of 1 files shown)


condnsdmatters:
I prefer this fix over #269 since the change is localised, and doesnt change behaviour for end users who chose to build on this. We should fix this properly later.

PR #283, opened 2021-10-28
Update tty tools

heiner:


nle/scripts/read_tty.py
```diff
@@ -49,42 +73,129 @@ def getfile(filename):
...
-                    datetime.datetime.fromtimestamp(timestamp),
-                    color(channel, 2),
-                    color("{", 11),
-                    data,
-                    color("}", 11),
+                char, *_ = struct.unpack("<B", data)
+                data = chr(char).encode("ascii", "backslashreplace")
+                arrow = "->"
+
+            data = str(data)[2:-1]  # Strip b' and '
```
condnsdmatters, line 172: probably better to use a more 'proper' way of doing this that doesnt depend on how Python decides to represent a converted bytes to string?

condnsdmatters (approved):

condnsdmatters:
Im just testing it out and will add some screen shots to the PR

nle/scripts/read_tty.py
heiner: I agree! But what is the proper way of doing exactly that? It's not `data.encode("ascii", "backslashreplace")` ...

heiner: StackOverflow seems to suggest doing what we are doing here ... https://stackoverflow.com/questions/54889234/how-to-decode-a-string-representation-of-a-bytes-object

condnsdmatters:
Screenshot 2021-11-02 at 15 09 14

setup.py
```diff
@@ -92,6 +92,8 @@ def build_extension(self, ext):
         "nle-play = nle.scripts.play:main",
         "nle-ttyrec = nle.scripts.ttyrec:main",
         "nle-ttyplay = nle.scripts.ttyplay:main",
+        "nle-ttyplay2 = nle.scripts.ttyplay2:main",
```
condnsdmatters, line 95: the ttyplay2 file doesnt have a main function? or something?

I'm getti
Screenshot 2021-11-02 at 15 10 31
ng

condnsdmatters:
Screenshot 2021-11-02 at 15 12 30

setup.py
heiner: Good catch; fixed.

nle/scripts/read_tty.py
heiner: Looks like this is the way to do this ... https://stackoverflow.com/questions/16748083/suppress-print-without-b-prefix-for-bytes-in-python-3 and https://github.com/python/cpython/blob/e1b900247227dad49d8231f1d028872412230ab4/Objects/bytesobject.c#L1345 ...

PR #288, opened 2021-11-26
Add glyph2tile array.

heiner:
This is to prepare a tile-based view.

Also fix some of our workflows which caused spurious failures.


samvelyan (approved):
Looks good. Thanks @heiner!

Issue #287, opened 2021-11-24
Race condition in environment settings


heiner:
## 🐛 Bug

Currently, we set environment variables before init and before reset. In the case of `HACKDIR`, this is problematic as there's a race when several instances of NLE get reset or initialized simultaneously.

On top of that, using environment variables is a bit ugly in our case.

We should create an nle_settings struct that contains these and funnel them through.

heiner:
Fixed in #291.

PR #291, opened 2021-11-30
Add nle_settings to nledl interface, don't use environment variables

heiner:
This also gets rid of the need to write a file for WIZKIT.

Fixes the race condition in https://github.com/facebookresearch/nle/issues/287

### DIFF
--- a/include/nleobs.h
+++ b/include/nleobs.h
@@ -78,4 +78,17 @@ typedef struct {
 #endif
 } nle_seeds_init_t;
 
+typedef struct nle_settings {
+    /*
+     *  Path to NetHack's game files.
+     */
+    char hackdir[4096];
+    /*
+     * Bool indicating whether to spawn monsters randomly after every step
+     * with some probability (1 by def). For more info, see
+     * https://nethackwiki.com/wiki/Monster_creation#Random_generation
+     */
+    int spawn_monsters;
+} nle_settings;
+
 #endif /* NLEOBS_H */
--- a/include/nleobs.h
+++ b/include/nleobs.h
@@ -84,6 +84,7 @@ typedef struct nle_settings {
      */
     char hackdir[4096];
     char options[32768];
+    char wizkit[4096];
     /*
      * Bool indicating whether to spawn monsters randomly after every step
      * with some probability (1 by def). For more info, see
--- a/include/nleobs.h
+++ b/include/nleobs.h
@@ -83,6 +83,7 @@ typedef struct nle_settings {
      *  Path to NetHack's game files.
      */
     char hackdir[4096];
+    char options[32768];
     /*
      * Bool indicating whether to spawn monsters randomly after every step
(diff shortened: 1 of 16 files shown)


samvelyan (approved):
Nice feature! Looks neat.

include/nleobs.h
```diff
@@ -78,4 +78,19 @@ typedef struct {
 #endif
 } nle_seeds_init_t;
 
+typedef struct nle_settings {
+    /*
+     *  Path to NetHack's game files.
+     */
+    char hackdir[4096];
+    char options[32768];
+    char wizkit[4096];
```
condnsdmatters, line 87: don't need TERM?

win/rl/pynethack.cc
```diff
@@ -71,15 +72,21 @@ class Nethack
...
 
-    Nethack(std::string dlpath, std::string hackdir, bool spawn_monsters)
+    Nethack(std::string dlpath, std::string hackdir,
+            std::string nethackoptions, bool spawn_monsters)
         : dlpath_(std::move(dlpath)), obs_{}, spawn_monsters_(spawn_monsters)
     {
         if (hackdir.size() > sizeof(settings_.hackdir) - 1) {
             throw std::length_error("hackdir too long");
         }
+        if (nethackoptions.size() > sizeof(settings_.options)) {
```
condnsdmatters, line 83: is one of these off by one?

include/nleobs.h
condnsdmatters: just saw that below we always set to "ansi" in get env. is it worth including it here instead?

win/rl/pynethack.cc
```diff
@@ -261,6 +262,15 @@ class Nethack
         return static_cast<game_end_types>(obs_.how_done);
     }
 
+    void
+    set_wizkit(std::string wizkit)
+    {
+        if (wizkit.size() > sizeof(settings_.wizkit)) {
```
condnsdmatters, line 268: same viz sizes

condnsdmatters (approved):
generally lgtm and a good change; just one question on off by ones, and one comment on putting many options in one spot (ie TERM to the options struct, and simplifying get env function - so if future we only set options in one location.)

include/nleobs.h
heiner: We kind of require "ansi" as the value ever since dropping the dependency on ncurses, so any other user-supplied value would break it I think.

win/rl/pynethack.cc
heiner: Good question. The reason for the difference is that in [nle.c](https://github.com/facebookresearch/nle/blob/heiner/nledl-interface/src/nle.c#L173) we need to add a `/` at the end. This is perhaps a bit unclean, but reflects what NetHack itself does in a comparable situation.

heiner: But that `/` only needs to be added to HACKDIR.

PR #292, opened 2021-11-30
Add def_oc_syms to pynethack.cc

heiner:
This allows us to generate a names of glyphs as in https://github.com/NetHack/NetHack/issues/634#issuecomment-970783123. For example with this code: https://gist.github.com/c5117ed04851c50744779af81825c899


samvelyan (approved):
LGTM!

PR #303, opened 2022-01-25
Fix _underscore access, rename Nethack variable as `nethack`.

### LINKED ISSUE #273: Don't rely on access to `_underscored` variables
## 🐛 Bug

Some relation to #272: We access things like `env._actions` in a number of places. This breaks when using certain gym wrappers, and also is plain bad style. (Perhaps okayish in some tests, but even there better avoided.)

Refactor to not do that anymore.

heiner:
Fixes #273.

One reason this matters more than one would assume is that gym envs wrappers tend to forward non-underscore access but will typically fail with `._underscore` access. This happened e.g. for the 0.21 release which auto-wrapped every env.

Also a bit more cleanup.

### DIFF
--- a/nle/env/base.py
+++ b/nle/env/base.py
@@ -248,15 +248,13 @@ class NLE(gym.Env):
 
         if actions is None:
             actions = FULL_ACTIONS
-        self._actions = actions
+        self.actions = actions
 
         self.last_observation = ()
 
         try:
             if savedir is None:
                 self.savedir = None
-                self._stats_file = None
-                self._stats_logger = None
             elif savedir:
                 self.savedir = os.path.abspath(savedir)
                 os.makedirs(self.savedir)
@@ -312,7 +310,7 @@ This might contain data that shouldn't be available to agents."""
         else:
             ttyrec = None
 
-        self.env = nethack.Nethack(
+        self.nethack = nethack.Nethack(
             observation_keys=self._observation_keys,
             options=options,
             playername="Agent-" + self.character,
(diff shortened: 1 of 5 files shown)


nle/env/base.py
```diff
@@ -312,15 +310,15 @@ def __init__(
         else:
             ttyrec = None
 
-        self.env = nethack.Nethack(
+        self.nethack = nethack.Nethack(
```
condnsdmatters, line 313: this is a much better renaming, though im sure will annoy downstream users.

condnsdmatters: so funny how easy it is to ignore "call it what it is!"

condnsdmatters (approved):
No more `env.env`! 🎉

nle/env/base.py
heiner: That's true re: downstream users. Hence I at least included @samvelyan here.

heiner: We could include an alias.

Then again, this is easy to fix.

samvelyan: Thanks for flagging this @heiner. I'll incorporate this into MiniHack after NLE's next release.

heiner: Again, if it's easier we can add an alias.

samvelyan:
Seems like this PR brakes the `play.py` script which uses `env._actions`. This is a small fix.

heiner:
> Seems like this PR brakes the `play.py` script which uses `env._actions`. This is a small fix.

Good catch, thanks! Fixed in https://github.com/facebookresearch/nle/pull/308

nle/env/base.py
hyzhak: Hi team, do you know when there will be the next release of MiniHack? I'm trying to use recent nle because of issue fixed by https://github.com/facebookresearch/nle/pull/299

PR #304, opened 2022-01-25
Add option to save ttyrec files every "M" episodes (#260)

### LINKED ISSUE #256: Save ttyrecs every "M" episode
Currently the base environment seems to write a ttyrec at [every call of reset](https://github.com/facebookresearch/nle/blob/main/nle/env/base.py#L462) (if `savedir is not None`). This creates a lot of ttyrec files.

It may be useful to add a parameter (say `save_episode_cycle`) to determine when to save a ttyrec - the ttyrec in that line can be set to None for every other episode in the cycle

Would this be a PR you alls would be interested in?

heiner:
Taking over from https://github.com/facebookresearch/nle/pull/260. Fixes #256.

### DIFF
--- a/nle/agent/agent.py
+++ b/nle/agent/agent.py
@@ -73,6 +73,9 @@ parser.add_argument("--disable_cuda", action="store_true",
                     help="Disable CUDA.")
 parser.add_argument("--use_lstm", action="store_true",
                     help="Use LSTM in agent model.")
+parser.add_argument("--episode_save_cycle", default=1, type=int,
+                    metavar="N", help="Cycle of episodes to save ttyrec.")
+
 
 # Loss settings.
 parser.add_argument("--entropy_cost", default=0.0006,
@@ -153,7 +156,9 @@ def act(
     try:
         logging.info("Actor %i started.", actor_index)
 
-        gym_env = create_env(flags.env, savedir=flags.rundir)
+        gym_env = create_env(
+            flags.env, savedir=flags.rundir, episode_save_cycle=flags.episode_save_cycle
+        )
         env = ResettingEnvironment(gym_env)
         env_output = env.initial()
         agent_state = model.initial_state(batch_size=1)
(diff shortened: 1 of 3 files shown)


condnsdmatters:
> Downside of PR as-is: Just setting savedir no longer saves data, one needs to also set save_ttyrec_every to a nonzero value now.
> 
> Alternative: Default save_ttyrec_every to -1 which could mean "1" if savedir is not-None, "0" otherwise?

Could we not just have a setting where if savedir is set - then we save every x. So the off switch is controlled by savedir, and there is no second off switch.

heiner:
> > Downside of PR as-is: Just setting savedir no longer saves data, one needs to also set save_ttyrec_every to a nonzero value now.
> > Alternative: Default save_ttyrec_every to -1 which could mean "1" if savedir is not-None, "0" otherwise?
> 
> Could we not just have a setting where if savedir is set - then we save every x. So the off switch is controlled by savedir, and there is no second off switch.

Yes.

My alternative suggestion was along those lines but not exactly that. The reason I went for something else first is that savedir has this additional twist where None means don't save, empty string means create a new random directory. I thought we could make the off switch the other flag instead...

condnsdmatters:
> > > Downside of PR as-is: Just setting savedir no longer saves data, one needs to also set save_ttyrec_every to a nonzero value now.
> > > Alternative: Default save_ttyrec_every to -1 which could mean "1" if savedir is not-None, "0" otherwise?
> > 
> > 
> > Could we not just have a setting where if savedir is set - then we save every x. So the off switch is controlled by savedir, and there is no second off switch.
> 
> Yes.
> 
> My alternative suggestion was along those lines but not exactly that. The reason I went for something else first is that savedir has this additional twist where None means don't save, empty string means create a new random directory. I thought we could make the off switch the other flag instead...

From a linguistic point of view, i think this PR is right to make 'save_ttyrec_every=0` be OFF, rather than `savedir=None`, and its cleaner to remove te `""` `None` case difference. I'm happy with this (breaking!) change

condnsdmatters (approved):

PR #305, opened 2022-01-26
Don't allow set_buffers to be called after reset.

### LINKED ISSUE #133: Reconsider set_buffers and observation keys.
Currently `set_buffers` is part of an API that is callable at any time. This means that in theory, the buffers to which observations are written/read from can be changed mid environment run. It also means more observation types can be added mid run.

Maintaining this feature requires storing intermediate results in an intermediate class and writing to observation buffers once - something that is actually violated already by both the checks for `nle_get_obs()->screen_description` visible in PR #131, and in writing directly to the `observation` in PR #119 . However, both these violations are for important performance reasons.

We should consider making the `set_buffers` method private at all times, and ensuring observation keys and buffers can only be set once at creation. This would allow us to write directly to the observation buffer further down the line, which in turn should give us a performance boost.

heiner:
This guarantees we write to the buffer from before and therefore
need to only update the delta, saving us a few memcopies potentially.

Fixes #133.

### DIFF
--- a/nle/tests/test_nethack.py
+++ b/nle/tests/test_nethack.py
@@ -200,6 +200,12 @@ class TestNetHackFurther:
         with pytest.raises(IOError):
             game.reset("")
 
+    def test_set_buffers_after_reset(self):
+        game = nethack.Nethack()
+        game.reset()
+        with pytest.raises(RuntimeError, match=r"set_buffers called after reset()"):
+            game._pynethack.set_buffers()
+
 
 class TestNethackSomeObs:
     @pytest.fixture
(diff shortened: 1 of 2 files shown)


condnsdmatters:
> This guarantees we write to the buffer from before and therefore need to only update the delta, saving us a few memcopies potentially.
> 
> Fixes #133.

Note for posterity - we do this in the tty observations where the callback from the virtual terminal on updating part of the screen involves writing to the observation.
(cf -  https://github.com/facebookresearch/nle/blob/main/src/nle.c#L69-L118)

condnsdmatters (approved):

heiner:
> Note for posterity - we do this in the tty observations where the callback from the virtual terminal on updating part of the screen involves writing to the observation.
> (cf - https://github.com/facebookresearch/nle/blob/main/src/nle.c#L69-L118)

Yes -- and specifically for the tty_chars observation it makes a lot of sense to only update the actually changed characters, like the actual tty would do. We also only write tty_cursor if it changed. Calling `set_buffers` after reset() was always broken in those cases (but also never part of our API anyway).

PR #307, opened 2022-01-26
Pick random race/gender/alignment when selecting random role.

### LINKED ISSUE #282: "Random" selection doesn't work the way it's intended
## 🐛 Bug

Right now, we sometimes use `name:Agent-@`, but that selects only a random _role_, not a random "race", "gender", or "alignment". NetHack has a flag called `flags.randomall`, enabled via the `-@` command line option, to truly select everything randomly.

We might want to expose the `flags` setting of NLE and enable that? It's kind of bad that there's no direct option for that. One alternative would be `NETHACKOPTIONS='name:Blob-@, race:random, gender:random, align:random'`.

heiner:
A NetHack option like "name:Heiner-@" only selects a random role but
still asks about race/gender/alignment. NetHack also has the command
line flag -@ (mapped to flags.randomall) to randomize everything,
but there's no option in the .nethackrc sense for that.

This change makes it such that playername="Heiner-@" appends
race:random, gender:random, align:random to NetHack's options unless
these have been set already.

Note: In the NetHack competition, this issue didn't hurt us because
we auto-selected "Random" for the startup questions.

Fixes #282.

### DIFF
--- a/nle/nethack/nethack.py
+++ b/nle/nethack/nethack.py
@@ -192,7 +192,7 @@ class Nethack:
         if options is None:
             options = NETHACKOPTIONS
         self.options = list(options) + ["name:" + playername]
-        if playername.split("-", 1)[1] == "@":
+        if next(iter(playername.split("-", 1)[1:]), None) == "@":
             # Random role. Unless otherwise specified, randomize
             # race/gender/alignment too.
             for key in ("race", "gender", "align"):
--- a/nle/nethack/nethack.py
+++ b/nle/nethack/nethack.py
@@ -192,7 +192,7 @@ class Nethack:
         if options is None:
             options = NETHACKOPTIONS
         self.options = list(options) + ["name:" + playername]
-        if next(iter(playername.split("-", 1)[1:]), None) == "@":
+        if "-" in playername and playername.split("-", 1)[1] == "@":
             # Random role. Unless otherwise specified, randomize
             # race/gender/alignment too.
             for key in ("race", "gender", "align"):
--- a/nle/nethack/nethack.py
+++ b/nle/nethack/nethack.py
@@ -192,7 +192,7 @@ class Nethack:
         if options is None:
             options = NETHACKOPTIONS
         self.options = list(options) + ["name:" + playername]
-        if "-" in playername and playername.split("-", 1)[1] == "@":
+        if playername.split("-", 1)[1:] == ["@"]:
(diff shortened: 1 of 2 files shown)


nle/nethack/nethack.py
```diff
@@ -192,6 +192,13 @@ def __init__(
         if options is None:
             options = NETHACKOPTIONS
         self.options = list(options) + ["name:" + playername]
+        if next(iter(playername.split("-", 1)[1:]), None) == "@":
```
condnsdmatters, line 195:
```suggestion
        if playername.endswith("-@"):
```

condnsdmatters (changes requested):

nle/nethack/nethack.py
condnsdmatters: i know the logics not identical, but convince me this wont do

heiner: I'd like to guard against situations like "Agent-bla-@". I've made that line simpler by doing `if "-" in playername` instead.

nle/nethack/nethack.py
```diff
@@ -192,6 +192,13 @@ def __init__(
...
+            # Random role. Unless otherwise specified, randomize
+            # race/gender/alignment too.
+            for key in ("race", "gender", "align"):
+                if not next((o for o in options if o.startswith(key + ":")), False):
```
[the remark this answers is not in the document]
heiner: Moved to the (in fact, much better) `if not any(o for o in options if o.startswith(key + ":")):`. Thanks for your insistence :)

nle/nethack/nethack.py
```diff
@@ -192,6 +192,13 @@ def __init__(
         if options is None:
             options = NETHACKOPTIONS
         self.options = list(options) + ["name:" + playername]
+        if "-" in playername and playername.split("-", 1)[1] == "@":
```
condnsdmatters, line 195:
```suggestion
        if playername.split("-", 1)[1:] == ["@"]:
```

condnsdmatters (approved):
one more suggestion but happy to accept now

nle/nethack/nethack.py
heiner: Now changed to your suggestion of `if playername.split("-", 1)[1:] == ["@"]:`

nle/nethack/nethack.py
heiner: Done in f231b99.

PR #315, opened 2022-04-14
Eric/add converter

condnsdmatters:
This is the state of the internal `main` branch when Heiner left. Everything so far has been reviewed internally.

The key feature here is to add a ttyrec converter in C++.

It involves committing the a version of libtmt, which has been modified, and undergoes some changes.

See commits for more details.


tscmoo (approved):
great, awesome, looks good!

drothermel:
So many tests, this looks great!

PR #316, opened 2022-04-14
Add an xlogfile to the generation of the ttyrecs.

condnsdmatters:
An xlogfile is a logfile that gives details of the episodes that have
just passed (https://nethackwiki.com/wiki/Xlogfile).  These were
previously being generated in the HACKDIR (temporary var dir). By
setting the SCOREPREFIX and touching the file, you can can change the
path that the file (always ending 'xlogfile') is generated.

NB: There is a lock based on fnctl on the logfile. To avoid multiple
nles in different trying to write to the same file (with the same savedir)
we prepend the process number, like ttyrecs. For writing to same savedir
within same process, an id should be added to both ttyrec and xlogfile.
This can be done in future (along with relabelling ttyrec to ttyrec2).

The log file records all COMPLETED episodes (in the NetHack sense),
since it derives from the features used to calculate top ten. In
standard usage it should lead to one row per episode, with all ttyrecs
from the same process in the same file. You can then find the ttyrec
corresponding to each row (start counting at 0th row).

### DIFF
--- a/include/nleobs.h
+++ b/include/nleobs.h
@@ -85,6 +85,7 @@ typedef struct nle_settings {
      *  Path to NetHack's game files.
      */
     char hackdir[4096];
+    char scoreprefix[4096];
     char options[32768];
     char wizkit[4096];
     /*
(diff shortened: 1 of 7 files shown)


condnsdmatters:
(Was review by heiner internally)

tscmoo (approved):

PR #318, opened 2022-04-26
Eric/dataset

condnsdmatters:


nle/dataset/dataset.py
```diff
@@ -128,7 +128,7 @@ def _ttyrec_generator(
         key_vals = [
             ("tty_chars", chars),
             ("tty_colors", colors),
-            ("tty_cursors", cursors),
+            ("tty_cursor", cursors),
```
drothermel: why singular if the variable name is plural?

nle/dataset/populate_db.py
```diff
@@ -50,6 +50,8 @@ def altorg_filename_to_timestamp(filename):
 
 
 def assign_ttyrecs_to_games(ttyrecs, games):
+    """Algorithm to assig a players ttyrecs to their games, knowing that only one game
```
drothermel: Assign

nle/dataset/populate_db.py
```diff
@@ -84,17 +89,51 @@ def assign_ttyrecs_to_games(ttyrecs, games):
...
+        ├── xlogfile.1
+        ├── xlogfile.2
+        ...
+        ├── about.txt
+        └── blacklist.txt
+
+    Note that unlike `nle` ttyrecs, altorg episodes may be split into parts.
+    We use a simple algorith based on the file creation times and xlogfile times
+    for the start and end of games to try to assign ttyrecs to games, knowing
+    a player can only ever be playing on game on altorg at a time.
```
drothermel, line 117: one game

nle/dataset/populate_db.py
```diff
@@ -153,38 +199,64 @@ def add_altorg_directory(path, name, filename=db.DB):
...
             insert_sql = f"""
                 INSERT INTO games
                 VALUES (NULL, {','.join('?' for _ in XLOGFILE_COLUMNS)} )
             """
             c.executemany(insert_sql, game_gen)
-
             gameids = db.get_most_recent_games(c.rowcount, conn=c)
             db.add_games(name, *gameids, conn=conn, commit=False)
 
+            # 4. Add ttyrecs to `ttyrecs` table.
```
drothermel, line 269: Does it make sense to allow the "blacklist" ability for these datasets too, even though we don't currently need it (for consistency)?

nle/dataset/db.py
```diff
@@ -228,7 +229,7 @@ def create(filename=DB):
             (
                 gameid        INTEGER,
                 dataset_name  TEXT,
-                PRIMARY KEY (gameid, dataset_name)
+                PRIMARY KEY   (dataset_name, gameid)
```
drothermel, line 232: Out of curiosity, why the swap from (gameid, dataset_name)?

drothermel (commented):
So many tests & comments, this looks great to me so far (I just left a few mainly tiny spelling suggestions).

nle/dataset/dataset.py
condnsdmatters: singular because we chose the same keys as NLE. There we have 1 frame of 80 x 24 chars, 1 frame of 80 x 24 colors, and 1 cursor (2 coords). In this case we technically have a batch of these. For consistency, it might be better to move to `timestamp` and `gameid` for the other keys.

nle/dataset/dataset.py
```diff
@@ -185,6 +183,7 @@ def __init__(
 
             for row in c.execute(self.sql_subset, (dataset_name,)).fetchall():
                 # A row is made up of [ gameid, part, path, meta1, meta2... metaN].
+                # if row[0] in gameids:
```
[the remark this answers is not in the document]
condnsdmatters: must delete

nle/dataset/populate_db.py
condnsdmatters: Probably not I think. YAGNI. This is because alt.org is truly a noisy dataset. If there are requests we can add later

nle/dataset/db.py
condnsdmatters: for multicolumn indices the order is important. sqlite creates  sorts the table on the first value, and then sorts duplicated values by the second column. in this case (clearly) in most cases, the dataset_name is what we are searching for, then maybe gameid.

heiner (approved):

nle/scripts/read_tty.py
```diff
@@ -181,6 +181,11 @@ def main():
                 char, *_ = struct.unpack("<B", data)
                 data = chr(char).encode("ascii", "backslashreplace")
                 arrow = "->"
+            elif channel == 2:
+                score, *_ = struct.unpack("<i", data)
+                # data = chr(score).encode("ascii", "backslashreplace")
```
drothermel, line 186: to delete?

PR #319, opened 2022-05-12
Add the filepath of the ttyrec if it exists to the xlogfile.

condnsdmatters:
This will make it easier for users to find which ttyrecs correspond to
which metadata.  Xlogfiles are only written to with natural endings; not
guaranteed if you call "reset" before "done" is ready. This means if
there is a panic and drop a game, you will not get alignment between
your ttyrec reset number and xlogfile row. This adds an explicit link.

NB: This only works because in nle (vs typical NetHack), we play the
whole game in _one_ ttyrec.

### DIFF
--- a/win/rl/pynethack.cc
+++ b/win/rl/pynethack.cc
@@ -341,6 +341,9 @@ class Nethack
     {
         py::gil_scoped_release gil;
 
+        if (!ttyrec)
+            strncpy(settings_.ttyrecname, "", sizeof(settings_.ttyrecname));
+
         if (!nle_) {
             nle_ =
                 nle_start(dlpath_.c_str(), &obs_, ttyrec ? ttyrec : ttyrec_,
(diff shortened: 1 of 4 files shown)


win/rl/pynethack.cc
```diff
@@ -122,6 +122,10 @@ class Nethack
         }
         strncpy(settings_.scoreprefix, scoreprefix.c_str(),
                 scoreprefix.length());
+        std::size_t found = ttyrec.rfind("/");
+        if (found!=std::string::npos && found + 1 < ttyrec.length())
```
condnsdmatters, line 126: why the `found+1` check? `found` points to the "/", so in case someone names their ttyrec "/path/ttyrec.bz2/". Highly unlikely - maybe it would break? Will check.

samvelyan (approved):
looks good

tscmoo (approved):

PR #320, opened 2022-05-16
Fixup libtmt to ensure that unrecognised terminal commands are ignored,

condnsdmatters:
to stop random letters writing to the screen.

Libtmt is a VT100 terminal emulator. We have adjusted it to let it
accept DEC and IBM graphical sets, however we don't intend to support
every terminal.  If a control sequence is not recognised (eg `ESC [ 1 ;
2; 1 z`  -- a command encountered alot in various ttyrecs) we should
abort the processing of the sequence rather than returning allowing the
`z` to be printed to the screen.


samvelyan (approved):
LGTM

PR #321, opened 2022-05-16
Populate database using the files in xlogfile

condnsdmatters:
ttyrecname=<filename> when writing ttyrecs.


condnsdmatters:
Prior to #319 , we populated the dataset's database by matching up rows in the xlogfile to the the `reset` suffix in `nle.<process>.<reset>.ttyrec<version>.bz2`.

This was clearly prone to mistakes when an episode was terminated (eg: hitting `reset()` before `done` is `True`), which would introduce offset errors. Now that the file has been added to the xlogfile, we can directly use this information to populate the db, removing this error.

drothermel (approved):
Much cleaner, nice!

PR #324, opened 2022-06-09
Adjust logging for populate db.

condnsdmatters:
We adjust the logging for db.py to be with a specific logger, and
populate_db methods now print. This is because populate methods should
happen very infrequently and should be logged loudly.


drothermel (approved):
Makes sense!

PR #326, opened 2022-06-09
Pin gym to 0.23.0 temporarily.

condnsdmatters:
NLE's interface does not pass new gym 0.24.0s checks. We temporarily
pin version while fix is in place.


drothermel (approved):
cool

PR #328, opened 2022-06-09
Add an `exists` method

condnsdmatters:
Maybe this is overkill, but i think it enables a good idiom:

```python
import nle.dataset as nld

if not nld.db.exists():
    nld.db.create()
    nld.add_nledata_directory(path_to_dir, name)
```


nmurraygit (approved):
Reasonable os.path check added.

PR #329, opened 2022-06-09
Rename "actions" to "keypresses" for dataset, which is more accurate.

condnsdmatters:


condnsdmatters:
If this doesnt break any tests, such a change definitely should ;) TBD.

minqi (approved):
lgtm

PR #330, opened 2022-06-16
Set the terminal emulator to by default not wrap extra long lines,

condnsdmatters:
thereby taking a crop of the larger terminal screen.


condnsdmatters:
This was tested manually:
**BEFORE**
Screenshot 2022-06-16 at 14 48 15


**AFTER**
Screenshot 2022-06-16 at 14 46 20

condnsdmatters:
> I think this is what the distinction of rendered terminal size vs cropped terminal size was for? But this here works too.

This is precisely what that was for (I mentioned this to @tscmoo), but this method is a little more memory efficient, and involves fewer parameters.

PR #334, opened 2022-08-04
Test and fix for issue 333.

### LINKED ISSUE #333: Bus Error Entering DLVL 10-12
## 🐛 Bug

Something about the NLE hits a bus error some of the time that the agent enters a floor between 10 and 12 of the main dungeon, or 7-9 in dungeon 2 (which I think is the Gnomish Mines?)

## To Reproduce

1. Go to nle/nle/env/tasks.py and temporarily comment out line 327, the line that prevents a NetHackChallenge env from being created with wizard mode enabled.
2. Run [the following code](https://github.com/facebookresearch/nle/files/9236546/sample.py.zip) and watch it crash with a bus error:

```
#!/usr/bin/env python3
import nle
import gym
import aicrowd_gym

keyLookup = {
	"a" : 24,
	"b" : 6,
	"c" : 30,
	"d" : 33,
	"e" : 35,
	"f" : 40,
	"g" : 72,
	"h" : 3,
	"i" : 44,
	"j" : 2,
	"k" : 0,
	"l" : 1,
	"m" : 54,
	"n" : 5,
	"o" : 57,
	"p" : 60,
	"q" : 64,
	"r" : 67,
	"s" : 75,
	"t" : 91,
	"u" : 4,
	"v" : 98,
	"w" : 102,
	"x" : 87,
	"y" : 7,
	"z" : 104,
	"A" : 89,
	"B" : 14,
	"C" : 27,
	"D" : 34,
	"E" : 36,
	"F" : 39,
	"G" : 73,
	"H" : 11,
	"I" : 45,
	"J" : 10,
	"K" : 8,
	"L" : 9,
	"M" : 55,
	"N" : 13,
	"O" : 58,
	"P" : 63,
	"Q" : 66,
	"R" : 69,
	"S" : 74,
	"T" : 88,
	"U" : 12,
	"V" : 43,
	"W" : 99,
	"X" : 95,
	"Y" : 15,
	"Z" : 28,
	"." : 18,
	"," : 61,
	"<" : 16,
	">" : 17,
	":" : 51,
	"0" : 110,
	"1" : 111,
	"2" : 112,
	"3" : 113,
	"4" : 114,
	"5" : 115,
	"6" : 116,
	"7" : 117,
	"8" : 118,
	"9" : 119,
	"$" : 120,
	"+" : 105,
	"-" : 106,
	" " : 107,
	"*" : 76,
	"#" : 20,
	"~" : 19, # represents enter
	"\\" : 38 # represents escape
}

env = aicrowd_gym.make("NetHackChallenge-v0", wizard=True, savedir=None)

sequence = "#wizlevelport~10~"

for x in range(1000):
	print(x)
	env.reset()
	for y in sequence:
		env.step(keyLookup[y])
```

## Environment

NLE version: 0.8.1
PyTorch version: 1.11.0
Is debug build: No
CUDA used to build PyTorch: None

OS: Mac OSX 12.1
GCC version: Could not collect
CMake version: Could not collect

Python version: 3.8
Is CUDA available: No
CUDA runtime version: No CUDA
GPU models and configuration: No CUDA
Nvidia driver version: No CUDA
cuDNN version: No CUDA

Versions of relevant libraries:
[pip3] msgpack-numpy==0.4.8
[pip3] numpy==1.22.4
[pip3] torch==1.11.0
[conda] Could not collect

## Additional context

If there's any more information you need, I'll try my best to provide it, but my ability to troubleshoot this problem is very limited. The bus error crashes my debugger without giving me a stack trace or anything, and execution seems to get passed to _pynethack.cpython-38-darwin.so, which is not a file format my IDE supports.

heiner:
Fixes #333.

### DIFF
--- a/src/nle.c
+++ b/src/nle.c
@@ -19,7 +19,7 @@
 #include <bzlib.h>
 #endif
 
-#define STACK_SIZE (1 << 15) /* 32KiB */
+#define STACK_SIZE (1 << 16) /* 64KiB */
 
 #ifndef __has_feature
 #define __has_feature(x) 0 /* Compatibility with non-clang compilers. */
--- a/nle/tests/test_envs.py
+++ b/nle/tests/test_envs.py
@@ -129,6 +129,27 @@ class TestGymEnv:
             assert "playmode:debug" not in env.nethack.options
 
 
+class TestWizardMode:
+    def test_wizlevelport(self):
+        actions = (
+            list(nethack.USEFUL_ACTIONS)
+            + list(nethack.TextCharacters)
+            + list(nethack.WizardCommand)
+        )
+
+        env = gym.make(
+            "NetHack-v0",
+            wizard=True,
+            actions=actions,
+            allow_all_yn_questions=True,
+            allow_all_modes=True,
+        )
+        env.reset()
+        env.step(actions.index(nethack.WizardCommand.WIZLEVELPORT))
+        for c in b"10\r":
+            env.step(actions.index(c))
+
+
 class TestWizkit:
     @pytest.fixture(autouse=True)  # will be applied to all tests in class
     def make_cwd_tmp(self, tmpdir):


heiner:
Looks like your flake8 and various tests are broken anyway. Proper FB engineering, here.

condnsdmatters:
> Looks like your flake8 and various tests are broken anyway. Proper FB engineering, here.

Rude.

heiner:
> > Looks like your flake8 and various tests are broken anyway. Proper FB engineering, here.
> 
> Rude.

:(

condnsdmatters (approved):

src/nle.c
```diff
@@ -19,7 +19,7 @@
 #include <bzlib.h>
 #endif
 
-#define STACK_SIZE (1 << 15) /* 32KiB */
+#define STACK_SIZE (1 << 16) /* 64KiB */
```
heiner, line 22: Perhaps a more fancy solution would be to do what pthreads do -- define a really large upper limit (8MiB or so) but use mmap to allocate the buffer and rely on page faults to grow actual memory usage. Perhaps @tscmoo can look into that :D

condnsdmatters:
Alles gut!

src/nle.c
heiner: I guess what could be done is to add a guard page at the end, a la https://man7.org/linux/man-pages/man3/pthread_attr_setguardsize.3.html

heiner: The above was in fact not that wrong after all. All ways of getting buffers short of `mmap` may/do initialize the memory and therefore page everything in.

But it turns out fcontext does exactly the right thing anyway: [It`mprotect`s the bottom page of the new stack.](https://github.com/septag/deboost.context/blob/325fed7ee8b760e4cb79030a58dab74821fdcba9/source/stack.c#L137) I don't quite understand why this is a SIGBUS then, not a SIGSEGV.

However, I did learn: The page size of M1 MacBooks is 16384 and our 32KiB value was two pages, one of them protected and unavailable. Also, `create_fcontext_stack` actually picks a sensible default of called with `0`, which we should just do.

heiner: Except that that default is currently broken due to a recent glibc change, reported at https://github.com/septag/deboost.context/issues/4.

PR #336, opened 2022-08-06
Use sensible fcontext default stack.

heiner:


heiner:
The test failure on all Linuxes is related to https://github.com/septag/deboost.context/issues/4.

condnsdmatters:
While the default stack method seems like a saner way to do things, is a resolution on this PR necessary?  I'd have thought unless we overflow again, probably not, but have not properly investigated this...  If so, shall I close and open an Issue?

heiner:
Answer to your question heavily depends on your definition of "necessary".

condnsdmatters:
lol i suppose so: i guess what i meant is - my understanding is that this PR aimed to use a default stack size instead of one arbititrarily chosen by us, but this is broken upstream. to my mind this PR is good practice but mainly cosmetic in that it shouldn't really affect most peoples running of NLE post pr #334 .  is there something i'm missing in the understanding of this issue that means NLE could break unexpectedly in future?

heiner:
No you're right. It could break if the stack gets very large but is unlikely to at this point. We can close this PR for now.

PR #350, opened 2022-12-14
Fix CI by fully specifying Python env and adjusting range on flakey

condnsdmatters:
test.


condnsdmatters:
This change emerged because ubuntu-latest on GH actions moved from 20.04 to 22.04 and there is no 3.6 available for it. We will keep on this version for now.

condnsdmatters:
See [here](https://github.com/actions/setup-python/issues/543)

samvelyan (approved):

PR #365, opened 2023-11-16
Fixup some CI

condnsdmatters:


condnsdmatters:
Looking at the Test & Deploys with cmate I think it might be some issue that gets fixed with an upgrade to gymnasium.

condnsdmatters:
Especially since the circl CI tests are done on 3.7.

samvelyan (approved):
LGTM.

